PR #111, opened 2025-08-03
Update README with Void client config and formatting

Froredion:
Added configuration instructions for the Void MCP client.


andrewpareles:
Thanks for this. This adds a ton of whitespace changes, can you only add the one change?

Froredion:
Woops I added it at 3 am, give me a sec

PR #184, opened 2025-08-26
add case study

zc277584121:


evaluation/case_study/django_14170/README.md
```diff
@@ -0,0 +1,128 @@
+# Django 14170: YearLookup ISO Year Bug
+
+A comparison showing how both methods(grep + semantic search) outperform grep-based approaches for complex Django ORM bugs.
```
codingjaguar, line 3: better call them:
Claude Context (semantic search + grep)
 Baseline (grep only)

codingjaguar: "Both Methods" sounds like two approaches rather than one composite approach

PR #194, opened 2025-09-03
AST PHP Support and Qdrant vector database support for a slimmer more streamlined setup

spale777:
Minimal core package changes only where necessary to allow for proper extension of the vectordb interface. 

Properly created and updated all relevant readme files. Created usage examples. Basic tests integrated and extended using Claude, all manually verified and all passing. 

Qdrant functionality matches closes to that of Milvus but much easier to maintain and deploy locally. Added some common and popular QWEN3 embedding models and custom configs for them, from my testing wouldn't go past 0.6B param one as i saw little to no improvement when using larger models but maybe there are some benefits there further testing is required.

Core change in terms of interfaces is around ID and query generation as those are database specific, was thinking of implementing a unified query format and then compile for each but proved to be too complex and not worth the effort we do two types of queries anyways at least for now.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,9 +32,12 @@ Model Context Protocol (MCP) allows you to integrate Claude Context with your fa
 ### Prerequisites
 
 <details>
-<summary>Get a free vector database on Zilliz Cloud 👈</summary>
+<summary>Get a vector database (Zilliz Cloud or Qdrant) 👈</summary>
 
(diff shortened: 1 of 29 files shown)


kierr:
This is a great contribution, well done! Especially nice to see Qwen3. You might consider breaking up the PR into multiple scopes though, if possible. For example, PHP vs Qdrant. It makes review and approval easier (and more likely!)

Keep in mind too, the authors of this project are Zilliz themselves :)

spale777:
Yeah in retrospect should have split it, but in general its like 30 lines of code for php support and qwen defaults.

PR #207, opened 2025-09-10
feat(core): Add PostgreSQL (with pgvector) as vector engine backend

YuweiXiao:
With this PR, users can now utilize claude-context in fully local manner, powered by a local Ollama embedding service and vector engine provided by a local PostgreSQL docker container.


tan-yong-sheng:
+1 hope for this support, thanks a lot for the work.

prom3theu5:
The work to support pgvector is great. IMO the data provider should have abstract implementations of each that way we can expand them out and rather than just marketing this locked to Milvus expand it out so the entire community can use it with whatever they want :)

no need to add the ollama backend code. Ollama already provides an openapi compatible api, and you can configure the base openapi url to use ollama or lm studio already

would be nice if we got some clear response as to yes or know if this will be considered for merging. If it is once I get some free time (not till after Christmas / feb) I’ll take it upon myself to implement chroma, pinecone and qdrant

my time is very limited atm unfortunately

YuweiXiao:
@prom3theu5 hi, thanks for taking a look. The repo’s been inactive for a month, so not sure if it’s still maintained. Ollama embeddings are already supported in the core package — I extended the code to make them configurable.

PR #261, opened 2026-02-01
Feature/azure ai search

svaus:


svaus:
(feat) added azure ai search, azure open ai and mcp http server support

PR #263, opened 2026-02-02
perf: cache compiled regex patterns in simpleGlobMatch

mareurs:
## Problem

The `simpleGlobMatch` function was creating a new `RegExp` object on every invocation. With large codebases (15k+ files) and many ignore patterns (300+), this caused millions of regex compilations per sync cycle, **pegging CPU at 100% indefinitely**.

### Root Cause Analysis

For a real-world example with 15,240 files and 332 ignore patterns:
- Each file requires ~5,000 `simpleGlobMatch` calls (nested loops check each path component against each pattern)
- Total: **35+ million regex compilations per sync**
- The MCP server would stay at 100% CPU forever, never completing the sync

## Solution

Added a `Map<string, RegExp>` cache for compiled regex patterns in both `Context` and `FileSynchronizer` classes. Since the same patterns are checked repeatedly, caching reduces regex compilations from millions to just the number of unique patterns (~300-600).

## Results

| Metric | Before | After |
|--------|--------|-------|
| CPU during sync | 100% (stuck) | Normal |
| Idle CPU after sync | 100% | 0% |
| Sync completion | Never | Completes normally |
| Sync time improvement | N/A | 30-70% faster |

## Changes

- `packages/core/src/context.ts`: Added `regexCache` property and caching logic in `simpleGlobMatch`
- `packages/core/src/sync/synchronizer.ts`: Same changes

## Testing

Tested with:
- Small codebases (~100 files): No noticeable difference
- Medium codebases (~1000 files): 30-50% faster sync
- Large codebases (15k+ files, 300+ ignore patterns): **Fixed infinite CPU loop**, sync now completes

### DIFF
--- a/packages/core/src/context.ts
+++ b/packages/core/src/context.ts
@@ -103,6 +103,7 @@ export class Context {
     private supportedExtensions: string[];
     private ignorePatterns: string[];
     private synchronizers = new Map<string, FileSynchronizer>();
+    private regexCache = new Map<string, RegExp>();
 
     constructor(config: ContextConfig = {}) {
         // Initialize services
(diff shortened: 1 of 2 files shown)


mareurs:
I still have issues with high CPU usage so let me test it some more to understand where the pb really is

mareurs:
The use was due to running node 22

PR #296, opened 2026-04-01
Fix snapshot 0/0 entries causing infinite force-reindex loop

### LINKED ISSUE #295: Snapshot 0/0 entries create infinite force-reindex loop
## Bug description

When a force reindex clears the Milvus collection, the snapshot file (`~/.context/mcp-codebase-snapshot.json`) is written with `indexedFiles: 0, totalChunks: 0`. If the subsequent indexing fails (network error, OOM, process killed, etc.), the snapshot is left at 0/0 permanently with status `completed` or `indexfailed`.

Every new Claude Code session sees the 0/0 entry, treats it as \"not indexed\", and the client passes `force: true` to `index_codebase`. This nukes the real Milvus collection (which may still have valid data) and writes 0/0 again — creating an infinite reindex loop.

## Reproduction steps

1. Index a codebase successfully (e.g. 13,000+ chunks)
2. Kill the MCP server process mid-reindex (or simulate a network failure to the embeddings server)
3. Observe `~/.context/mcp-codebase-snapshot.json` — the entry will show `indexedFiles: 0, totalChunks: 0`
4. Start a new Claude Code session in that project
5. The session calls `index_codebase` with `force: true` because the snapshot says 0/0
6. This nukes the Milvus collection and writes 0/0 again
7. Repeat forever

## Root cause

Three interacting issues:

### 1. Recovery path hardcodes 0/0

In `handlers.ts`, the search fallback recovery writes:
```typescript
this.snapshotManager.setCodebaseIndexed(absolutePath,
  { indexedFiles: 0, totalChunks: 0, status: completed as const });
```
It never queries Milvus for actual row counts.

### 2. No startup validation against Milvus

`loadCodebaseSnapshot()` in `snapshot.ts` loads the snapshot from disk as-is. The v0.1.5 fix (PR #282) resets `indexing` entries to `indexfailed`, but does NOT validate `completed` entries with 0/0 stats against the actual Milvus backend. There is no check that says \"this entry claims 0 files indexed — does the collection actually have data?\"

### 3. Force reindex writes 0/0 before indexing starts

In `handlers.ts`:
```typescript
this.snapshotManager.setCodebaseIndexing(absolutePath, 0);
this.snapshotManager.saveCodebaseSnapshot();
```
If indexing fails after this point, the snapshot has 0/0 and the v0.1.5 startup fix converts it to `indexfailed` — but the real collection data was already nuked by the force path.

## Observed behavior

Current snapshot on my system shows 4 of 5 codebases at 0/0 despite having real Milvus collections with data:

```
indexed    files=    0 chunks=    0  /home/eric/bu-aws          (Milvus: 13,395 rows)
indexed    files=    0 chunks=    0  /home/eric/ln-marketing-brief (Milvus: 23 rows)
indexed    files=    0 chunks=    0  /home/eric/page-toolkit     (Milvus: 474 rows)
indexed    files=    0 chunks=    0  /home/eric/wow-addons       (Milvus: 87 rows)
indexed    files=  918 chunks= 2525  /bulk-storage/repos/mission-control (OK)
```

Manually patching the snapshot to reflect real Milvus row counts fixes it temporarily, but the bug re-triggers within hours.

## Suggested fix

1. **Startup validation**: In `loadCodebaseSnapshot()`, for any entry with `status: completed` and `indexedFiles: 0, totalChunks: 0`, query Milvus `collection.stats()` for actual row count. If the collection exists with data, update the snapshot. If no collection exists, remove the entry.

2. **Recovery path should query stats**: Change the search fallback recovery from hardcoded `{ indexedFiles: 0, totalChunks: 0 }` to querying actual Milvus collection stats before writing the snapshot entry.

3. **Never write completed with 0/0**: `setCodebaseIndexed` should reject or warn when called with zero files — either get real stats or mark as needs-reindex.

## Environment

- `@zilliz/claude-context-mcp` v0.1.6 (latest as of 2026-04-01)
- Milvus 2.x (self-hosted via Docker)
- Self-hosted embeddings server
- Claude Code CLI

## Related issues

- #249 — Index not auto-loaded from vector DB on MCP server restart (closest to this bug)
- #276 — Snapshot file gets overwritten (fixed in v0.1.5, addresses multi-process overwrites but not 0/0 corruption)
- #257 — Orphaned merkle files not cleaned up on index clear (related cruft from force reindex)

EricSeastrand:
## Summary

Fixes #295 — snapshot recovery paths write \`indexedFiles: 0, totalChunks: 0\` without querying Milvus for actual stats, creating an infinite reindex loop.

**The bug:** When a force reindex clears data or recovery fires, the snapshot is written with 0/0 stats. If indexing subsequently fails, every new session sees 0/0 → treats it as unindexed → passes \`force: true\` → nukes the real Milvus data → writes 0/0 again.

**The fix (3 parts):**

1. **New \`getCollectionRowCount()\` method** on \`VectorDatabase\` interface + both implementations (SDK and REST). Queries actual Milvus collection statistics instead of hardcoding zeros.

2. **Fixed 3 recovery paths** in \`handlers.ts\` that hardcoded \`{ indexedFiles: 0, totalChunks: 0 }\`:
   - \`syncIndexedCodebasesFromCloud()\` — cloud recovery
   - \`handleIndexCodebase()\` — snapshot/VectorDB desync recovery
   - \`handleSearchCode()\` — search fallback recovery
   
   All three now call \`getCollectionRowCount()\` to get real stats from Milvus.

3. **New \`validateSnapshotZeroEntries()\` method** that runs at the start of index/search operations. For any snapshot entry with status=indexed but 0/0 stats:
   - If Milvus has data → updates snapshot with real row count
   - If Milvus has no collection → removes the stale entry

## Changes

| File | Change |
|------|--------|
| \`packages/core/src/vectordb/types.ts\` | Add \`getCollectionRowCount()\` to \`VectorDatabase\` interface |
| \`packages/core/src/vectordb/milvus-vectordb.ts\` | Implement via \`getCollectionStatistics()\` SDK method |
| \`packages/core/src/vectordb/milvus-restful-vectordb.ts\` | Implement via REST \`/v2/vectordb/collections/get_stats\` |
| \`packages/mcp/src/handlers.ts\` | Fix 3 recovery paths + add startup validation |

## Test plan

- [x] \`pnpm typecheck\` passes across all packages
- [x] \`pnpm build:core\` and \`pnpm build:mcp\` compile cleanly
- [ ] Manual test: corrupt snapshot to 0/0, verify recovery queries Milvus and restores real stats
- [ ] Manual test: remove Milvus collection but leave snapshot entry, verify stale entry is cleaned up

### DIFF
--- a/packages/core/src/vectordb/milvus-restful-vectordb.ts
+++ b/packages/core/src/vectordb/milvus-restful-vectordb.ts
@@ -823,4 +823,25 @@ export class MilvusRestfulVectorDatabase implements VectorDatabase {
         console.warn('[MilvusRestfulDB] ⚠️  checkCollectionLimit not implemented for REST API - returning true');
         return true;
     }
+
+    async getCollectionRowCount(collectionName: string): Promise<number> {
+        try {
+            const hasCol = await this.hasCollection(collectionName);
+            if (!hasCol) {
+                return 0;
+            }
+
+            const response = await this.makeRequest(
+                '/v2/vectordb/collections/get_stats',
+                'POST',
+                { collectionName }
+            );
+
+            const rowCount = response?.data?.rowCount ?? response?.data?.row_count;
+            return typeof rowCount === 'number' ? rowCount : parseInt(rowCount, 10) || 0;
+        } catch (error) {
+            console.error(`[MilvusRestfulDB] Error getting row count for '${collectionName}':`, error);
+            return 0;
+        }
+    }
 }
--- a/packages/core/src/vectordb/milvus-vectordb.ts
+++ b/packages/core/src/vectordb/milvus-vectordb.ts
@@ -781,4 +781,44 @@ export class MilvusVectorDatabase implements VectorDatabase {
             throw error;
         }
     }
+
+    async getCollectionRowCount(collectionName: string): Promise<number> {
+        await this.ensureInitialized();
+
+        if (!this.client) {
+            throw new Error('MilvusClient is not initialized after ensureInitialized().');
+        }
+
+        try {
+            const hasCol = await this.client.hasCollection({ collection_name: collectionName });
+            if (!hasCol.value) {
+                return 0;
+            }
+
+            const result = await this.client.getCollectionStatistics({
+                collection_name: collectionName,
+            });
+
+            if (result.status.error_code !== 'Success') {
+                console.warn(`[MilvusDB] Failed to get stats for '${collectionName}': ${result.status.reason}`);
+                return 0;
+            }
+
+            // row_count is returned in the stats data array
+            const rowCountStat = result.data?.rows_count ?? result.stats?.find((s: any) => s.key === 'row_count');
+            if (typeof rowCountStat === 'number') {
+                return rowCountStat;
+            }
+            if (rowCountStat && typeof rowCountStat === 'object' && 'value' in rowCountStat) {
+                return parseInt(rowCountStat.value, 10) || 0;
+            }
+
+            // Fallback: parse from data.row_count if available
+            const rc = (result as any).data?.row_count;
+            return typeof rc === 'number' ? rc : parseInt(rc, 10) || 0;
+        } catch (error) {
+            console.error(`[MilvusDB] Error getting row count for '${collectionName}':`, error);
+            return 0;
+        }
+    }
 }
--- a/packages/core/src/vectordb/types.ts
+++ b/packages/core/src/vectordb/types.ts
@@ -138,6 +138,14 @@ export interface VectorDatabase {
      * Returns true if collection can be created, false if limit exceeded
      */
     checkCollectionLimit(): Promise<boolean>;
+
+    /**
+     * Get the number of entities (rows) in a collection.
+     * Used by snapshot recovery to get actual stats instead of hardcoding zeros.
+     * @param collectionName Collection name
+     * @returns Number of entities in the collection, or 0 if collection doesn't exist
+     */
+    getCollectionRowCount(collectionName: string): Promise<number>;
 }
 
 /**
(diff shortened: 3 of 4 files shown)


EricSeastrand:
My Claude code got trigger happy and created this PR for me.
I have not verified it yet myself. I will update this comment if that changes. But I felt it helpful to leave for reference.

I can confirm that I am having this issue: I keep re-indexing the entire code base unnecessarily. I cannot confirm that anything in this PR is actually the root cause, much less that the PR actually addresses it.

Thanks for this great MCP tool!

Nottlespike:
> My Claude code got trigger happy and created this PR for me. I have not verified it yet myself. I will update this comment if that changes. But I felt it helpful to leave for reference.
> 
> I can confirm that I am having this issue: I keep re-indexing the entire code base unnecessarily. I cannot confirm that anything in this PR is actually the root cause, much less that the PR actually addresses it.
> 
> Thanks for this great MCP tool!

The current docs seems a bit underwhelming. I went ahead and made a PR specifically for this. I would mark this as a draft if you plan on fixing this.

EricSeastrand:
Closing as it was addressed in #303

zc277584121:
Thanks for filing the detailed issue (#295) and this PR — the root cause analysis was spot-on and directly informed the final fix.

This has been addressed in #303 (merged) and released as **v0.1.7**. The approach is similar to what you proposed here but with a few refinements discovered during E2E testing:

- Uses `count(*)` query instead of `getCollectionStatistics()` — we found that stats return stale zeros for unflushed data, causing false negatives even with the fix applied.
- Returns `-1` (not `0`) as the sentinel for "unknown" row count, so callers can distinguish "query failed" from "truly empty".
- Adds a startup `validateLegacyZeroEntries()` to auto-heal existing 0/0 entries on upgrade.
- Adds a defensive guard in `setCodebaseIndexed()` that rejects 0/0+completed writes at the data model level.

Feel free to close this PR. Thanks again for the thorough investigation!

PR #302, opened 2026-04-13
docs(mcp): explain indexing status semantics

Nottlespike:
## Summary
- document that `get_indexing_status` reports coarse phase-based progress rather than a live file-count ratio
- explain why completed entries can sometimes show stale `0 files, 0 chunks` snapshot metadata
- clarify that indexed codebases are keyed by absolute path, so the same repository under different paths is tracked separately

## Docs updated
- `packages/mcp/README.md`
- `docs/dive-deep/asynchronous-indexing-workflow.md`
- `docs/troubleshooting/faq.md`

## Verification
- `pnpm install`
- `pnpm build`

Both commands completed successfully in the local fork clone before opening this PR.

### DIFF
--- a/docs/dive-deep/asynchronous-indexing-workflow.md
+++ b/docs/dive-deep/asynchronous-indexing-workflow.md
@@ -35,6 +35,39 @@ The flow diagram above shows the complete indexing workflow, illustrating how th
 - **`indexfailed`** - ❌ Error occurred, can retry
 - **`not_found`** - ❌ Not indexed yet
 
+## How Progress Is Calculated
+
+`get_indexing_status` reports a **coarse, phase-based percentage**, not an exact fraction of files completed.
+
+- **0%** - Preparing the target collection and validating indexing prerequisites
+- **~5%** - Scanning the codebase and building the file list
+- **10% → 100%** - Processing files, chunking code, generating embeddings, and writing batches to the vector database
+- **100%** - Indexing finished successfully
+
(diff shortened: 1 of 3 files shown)


zc277584121:
Thanks for documenting this — the context around phase-based progress and absolute path keying is useful regardless.

However, the `0 files, 0 chunks` sections may need updating to reflect the fix in v0.1.7 (#303). As of v0.1.7:

1. **Recovery paths** no longer write hardcoded `{indexedFiles: 0, totalChunks: 0}` — they query Milvus for the real row count via `count(*)`, and skip the write entirely if the count is unknown or zero.
2. **Startup validation** (`validateLegacyZeroEntries`) auto-heals any legacy 0/0+completed entries on MCP server start — querying Milvus and either fixing the count or removing the stale entry.
3. **Defensive guard** in `setCodebaseIndexed` rejects 0/0+completed writes at the data model level, so no future code path can produce this state.

This means users on v0.1.7+ should essentially **never see `0 files, 0 chunks`** in `get_indexing_status`. The manual workaround ("clear and re-index") is no longer needed.

Suggested changes to the docs:

- The FAQ entry "Why does `get_indexing_status` show `0 files, 0 chunks`?" could be reframed as: "In versions prior to v0.1.7 this could happen due to stale snapshot metadata. As of v0.1.7, the MCP server automatically detects and repairs these entries on startup. If you still see 0/0 after upgrading, please file an issue."
- The inline note in `packages/mcp/README.md` ("If a completed entry shows `0 files, 0 chunks`, that usually means...") should be updated similarly.
- The progress explanation and absolute-path keying sections are great as-is — no changes needed there.

Happy to help refine the wording if needed.

PR #304, opened 2026-04-20
feat: add OpenRouter as embedding provider name

BeamNawapat:
## Summary

Add OpenRouter as a first-class provider name for cleaner UX. Reuses `OpenAIEmbedding` internally with OpenRouter's OpenAI-compatible endpoint — zero code duplication.

## Motivation

Users can achieve the same result via `OPENAI_BASE_URL` workaround, but a named provider gives better discoverability and simpler configuration.

## Changes (2 files, ~30 lines)

- `packages/mcp/src/config.ts` — Add `OpenRouter` to provider union, `OPENROUTER_API_KEY` env var, default model, help message
- `packages/mcp/src/embedding.ts` — Add `OpenRouter` case that creates `OpenAIEmbedding` with `baseURL: 'https://openrouter.ai/api/v1'`

No new files. No new dependencies. No code duplication.

## Usage

```bash
EMBEDDING_PROVIDER=OpenRouter \
OPENROUTER_API_KEY=sk-or-v1-xxx \
EMBEDDING_MODEL=openai/text-embedding-3-small \
npx @zilliz/claude-context-mcp@latest
```

## Test plan

- [ ] OpenRouter provider starts and connects
- [ ] Existing providers unaffected (no regression)
- [ ] `pnpm build` passes

### DIFF
--- a/packages/mcp/src/config.ts
+++ b/packages/mcp/src/config.ts
@@ -4,7 +4,7 @@ export interface ContextMcpConfig {
     name: string;
     version: string;
     // Embedding provider configuration
-    embeddingProvider: 'OpenAI' | 'VoyageAI' | 'Gemini' | 'Ollama';
+    embeddingProvider: 'OpenAI' | 'VoyageAI' | 'Gemini' | 'Ollama' | 'OpenRouter';
     embeddingModel: string;
     // Provider-specific API keys
     openaiApiKey?: string;
@@ -12,6 +12,8 @@ export interface ContextMcpConfig {
     voyageaiApiKey?: string;
     geminiApiKey?: string;
     geminiBaseUrl?: string;
+    // OpenRouter configuration
+    openrouterApiKey?: string;
     // Ollama configuration
     ollamaModel?: string;
     ollamaHost?: string;
@@ -76,6 +78,8 @@ export function getDefaultModelForProvider(provider: string): string {
             return 'voyage-code-3';
         case 'Gemini':
             return 'gemini-embedding-001';
+        case 'OpenRouter':
+            return 'openai/text-embedding-3-small';
         case 'Ollama':
             return 'nomic-embed-text';
         default:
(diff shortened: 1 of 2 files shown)


zc277584121 (changes requested):
Thanks for the contribution! The motivation makes sense — giving OpenRouter users a cleaner config experience. However, I have some concerns about the current approach.

## Main issue: code duplication

`openrouter-embedding.ts` (151 lines) is nearly identical to `openai-embedding.ts`. Both use the `openai` npm package, same `embeddings.create()` calls, same error handling, same `detectDimension` logic. The only real differences are:

- Default `baseURL` → `https://openrouter.ai/api/v1`
- Default model → `openai/text-embedding-3-small`
- Supported model list (3 entries, all just OpenAI models with `openai/` prefix)

Users can already achieve the same result today:
```bash
EMBEDDING_PROVIDER=OpenAI \
OPENAI_BASE_URL=https://openrouter.ai/api/v1 \
OPENAI_API_KEY=sk-or-v1-xxx \
EMBEDDING_MODEL=openai/text-embedding-3-small
```

## Suggested alternatives

If we do want a first-class `OpenRouter` provider name for better UX, it should reuse `OpenAIEmbedding` rather than duplicate it. Something like:

```typescript
// In createEmbeddingInstance:
case 'OpenRouter':
    if (!config.openrouterApiKey) {
        throw new Error('OPENROUTER_API_KEY is required');
    }
    return new OpenAIEmbedding({
        apiKey: config.openrouterApiKey,
        model: config.embeddingModel,
        baseURL: 'https://openrouter.ai/api/v1',
    });
```

This achieves the same user-facing behavior (~5 lines vs 151 lines) and avoids maintaining two copies of the same logic.

Alternatively, just document the `OPENAI_BASE_URL` approach in the README — that may be sufficient.

BeamNawapat:
Good call — 151 lines of duplication was unnecessary. Reworked to reuse `OpenAIEmbedding` directly:

```typescript
case 'OpenRouter':
    return new OpenAIEmbedding({
        apiKey: config.openrouterApiKey,
        model: config.embeddingModel,
        baseURL: 'https://openrouter.ai/api/v1',
    });
```

Force-pushed. Now 2 files changed, ~30 lines total. No new files, no new dependencies.

PR #305, opened 2026-04-20
feat: add VoyageAI custom base URL support

BeamNawapat:
## Summary

Add `VOYAGEAI_BASE_URL` environment variable to allow VoyageAI embedding provider to use custom API endpoints.

## Motivation

After MongoDB acquired VoyageAI, API keys created through MongoDB Atlas only authenticate against `https://ai.mongodb.com/v1/embeddings`, while keys created through VoyageAI directly only work with `https://api.voyageai.com/v1/embeddings`. The current implementation hardcodes the VoyageAI endpoint, causing 403 errors for MongoDB Atlas users.

When `baseURL` is set, the implementation uses the OpenAI SDK internally because MongoDB's VoyageAI endpoint only accepts OpenAI-compatible request format (different field naming from the native VoyageAI API). The native VoyageAI SDK's `environment` option changes the URL but not the request format, so it still gets 400 errors from MongoDB. This tradeoff is documented with inline comments.

## Changes

- `packages/core/src/embedding/voyageai-embedding.ts` — Add `baseURL` to config, use OpenAI SDK when `baseURL` is set (documented why), clean `getClient()` return type
- `packages/mcp/src/config.ts` — Add `voyageaiBaseUrl` config field, `VOYAGEAI_BASE_URL` env var mapping, debug dump, help message
- `packages/mcp/src/embedding.ts` — Pass `baseURL` when creating VoyageAI instance, log Base URL

## Usage

```bash
# MongoDB Atlas VoyageAI key
EMBEDDING_PROVIDER=VoyageAI \
VOYAGEAI_API_KEY=al-xxx \
VOYAGEAI_BASE_URL=https://ai.mongodb.com/v1 \
npx @zilliz/claude-context-mcp@latest

# Traditional VoyageAI key (backward compatible, no change needed)
EMBEDDING_PROVIDER=VoyageAI \
VOYAGEAI_API_KEY=pa-xxx \
npx @zilliz/claude-context-mcp@latest
```

## Note

Per review feedback, the following have been split into separate PRs:
- voyage-4 models + remove dead method → separate PR
- Dimension parsing regex → separate PR
- `input_type` query/document support → will redesign as parameter-based approach to avoid concurrency issues

## Test plan

- [ ] VoyageAI works without `VOYAGEAI_BASE_URL` (backward compatible)
- [ ] VoyageAI works with MongoDB Atlas endpoint and key
- [ ] `pnpm build` passes

### DIFF
--- a/packages/core/src/context.ts
+++ b/packages/core/src/context.ts
@@ -324,7 +324,7 @@ export class Context {
             await this.loadIgnorePatterns(codebasePath);
 
             // To be safe, let's initialize if it's not there.
-            const newSynchronizer = new FileSynchronizer(codebasePath, this.ignorePatterns);
+            const newSynchronizer = new FileSynchronizer(codebasePath, this.ignorePatterns, this.supportedExtensions);
             await newSynchronizer.initialize();
             this.synchronizers.set(collectionName, newSynchronizer);
         }
--- a/packages/core/src/sync/synchronizer.ts
+++ b/packages/core/src/sync/synchronizer.ts
@@ -10,13 +10,15 @@ export class FileSynchronizer {
     private rootDir: string;
     private snapshotPath: string;
     private ignorePatterns: string[];
+    private supportedExtensions: string[];
 
-    constructor(rootDir: string, ignorePatterns: string[] = []) {
+    constructor(rootDir: string, ignorePatterns: string[] = [], supportedExtensions: string[] = []) {
         this.rootDir = rootDir;
         this.snapshotPath = this.getSnapshotPath(rootDir);
         this.fileHashes = new Map();
         this.merkleDAG = new MerkleDAG();
         this.ignorePatterns = ignorePatterns;
+        this.supportedExtensions = supportedExtensions;
     }
 
     private getSnapshotPath(codebasePath: string): string {
@@ -81,6 +83,10 @@ export class FileSynchronizer {
             } else if (stat.isFile()) {
                 // Verify it's really a file and not ignored
                 if (!this.shouldIgnore(relativePath, false)) {
+                    const ext = path.extname(entry.name);
+                    if (this.supportedExtensions.length > 0 && !this.supportedExtensions.includes(ext)) {
+                        continue;
+                    }
                     try {
                         const hash = await this.hashFile(fullPath);
                         fileHashes.set(relativePath, hash);
--- a/packages/mcp/src/handlers.ts
+++ b/packages/mcp/src/handlers.ts
@@ -531,7 +531,7 @@ export class ToolHandlers {
             const { FileSynchronizer } = await import("@zilliz/claude-context-core");
             const ignorePatterns = this.context.getIgnorePatterns() || [];
             console.log(`[BACKGROUND-INDEX] Using ignore patterns: ${ignorePatterns.join(', ')}`);
-            const synchronizer = new FileSynchronizer(absolutePath, ignorePatterns);
+            const synchronizer = new FileSynchronizer(absolutePath, ignorePatterns, this.context.getSupportedExtensions());
             await synchronizer.initialize();
 
             // Store synchronizer in the context (let context manage collection names)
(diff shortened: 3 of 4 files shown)


zc277584121 (changes requested):
Thanks for this PR! The base URL support for MongoDB Atlas is a valid use case, and adding voyage-4 models is welcome. However, there are several issues I'd like to see addressed before merging.

## 1. Mixed concerns — please split into separate PRs

This PR bundles at least 4 unrelated changes:

1. **VoyageAI custom base URL** (the core feature)
2. **voyage-4 model list** (independent data addition)
3. **`setMode()` on base Embedding class + context.ts search changes** (unrelated feature)
4. **Remove duplicate `updateDimensionForModel`** (refactoring)

Each of these should be a separate PR. The voyage-4 models and the `updateDimensionForModel` cleanup are straightforward and can be merged quickly on their own. The `setMode()` changes need design discussion (see below).

## 2. `setMode()` has a concurrency issue

The pattern in context.ts:

```typescript
this.embedding.setMode('query');
try {
    queryEmbedding = await this.embedding.embed(query);
} finally {
    this.embedding.setMode('document');
}
```

This mutates shared state on the embedding instance. If two search requests arrive concurrently, they'll interfere with each other's input type (request A sets 'query', request B sets 'query', A finishes and resets to 'document', B's embed call now runs with 'document' mode).

A safer design would be to pass the mode as a parameter to `embed()` / `embedBatch()`, e.g.:

```typescript
embed(text: string, options?: { inputType?: 'document' | 'query' }): Promise<EmbeddingVector>
```

Also, `setMode()` as a no-op on the base `Embedding` class is an abstraction leak — `input_type` is VoyageAI-specific. Other providers don't have this concept and shouldn't carry a no-op method for it.

## 3. OpenAI SDK fallback — `@ts-ignore` and breaking changes

When `baseURL` is set, the implementation switches from `VoyageAIClient` to `OpenAI` SDK entirely. This causes:

- `client` becomes nullable (`VoyageAIClient | null`) — adds null checks everywhere
- `// @ts-ignore` on `input_type` in the OpenAI path
- `getClient()` return type changes from `VoyageAIClient` to `VoyageAIClient | OpenAI | null` — this is a breaking change for any downstream code using `getClient()`

If the VoyageAI SDK's `environment` option can't be used (because MongoDB's endpoint requires OpenAI-compatible format), that tradeoff is understandable, but please document it with a comment explaining *why* the OpenAI SDK is needed instead of the native one.

## What I'd suggest

1. **Split out voyage-4 models + `updateDimensionForModel` cleanup** into a small PR — can merge immediately.
2. **Split out dimension parsing improvement** (regex from string) — also straightforward.
3. **Base URL support** as its own PR, with the `getClient()` typing cleaned up and the reason for OpenAI SDK fallback documented.
4. **`setMode()` / query-vs-document input type** as a separate PR with a design that avoids mutable state (pass as parameter, not setter).

BeamNawapat:
Thanks for the thorough review! All points are valid. Here's what I've done:

**Split into separate PRs as suggested:**

1. **voyage-4 models + remove dead `updateDimensionForModel`** → `feat/voyage-4-models` (ready for review)
2. **Dimension parsing regex** → `feat/dimension-parsing` (ready for review)
3. **This PR (#305)** — now contains only VoyageAI base URL support
   - Documented why OpenAI SDK fallback is needed (MongoDB endpoint only accepts OpenAI-compatible format)
   - Cleaned up `getClient()` → returns `VoyageAIClient | null`, added `isOpenAICompatMode()` for downstream introspection
   - Removed `@ts-ignore` — using explicit `Record<string, unknown>` cast instead
4. **input_type query/document** — will submit as separate PR with parameter-based design per your suggestion:
   ```typescript
   embed(text: string, options?: { inputType?: 'document' | 'query' }): Promise<EmbeddingVector>
   ```
   This avoids mutable shared state and keeps `input_type` within VoyageAI's implementation without leaking to the base `Embedding` class.

Force-pushed this branch to contain only the base URL changes (1 commit).

PR #306, opened 2026-04-21
feat: add voyage-4 model series and remove unused method

BeamNawapat:
## Summary

- Add voyage-4-large, voyage-4, voyage-4-lite, and voyage-4-nano to the VoyageAI supported models list (released January 2026)
- Remove unused `updateDimensionForModel()` which duplicated logic from `updateModelSettings()`

## Changes

- `packages/core/src/embedding/voyageai-embedding.ts` — Add 4 voyage-4 models, remove dead method

## Test plan

- [ ] `pnpm build` passes
- [ ] Existing VoyageAI models still work (no regression)

### DIFF
--- a/packages/core/src/embedding/voyageai-embedding.ts
+++ b/packages/core/src/embedding/voyageai-embedding.ts
@@ -44,23 +44,6 @@ export class VoyageAIEmbedding extends Embedding {
         }
     }
 
-    private updateDimensionForModel(model: string): void {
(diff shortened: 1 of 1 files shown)


zc277584121 (approved):
Clean split from #305 — exactly what was needed. The voyage-4 model data looks correct, and removing the duplicate `updateDimensionForModel` is a good cleanup since `updateModelSettings` already covers the same logic.

LGTM.

PR #314, opened 2026-04-22
feat(mcp): make background sync opt-in

### LINKED ISSUE #285: Reduce CPU churn from multiple MCP instances and support shared/remote deployment
## Problem

Running `@zilliz/claude-context-mcp` as a local MCP in tools like Opencode can cause high aggregate CPU usage when multiple workspaces/sessions are open at once.

In my case, each workspace spawned its own `claude-context-mcp` process, and each process appeared to:

- start background sync automatically on server startup
- perform an initial sync shortly after boot
- periodically rescan all indexed codebases from the shared snapshot
- contend on the shared snapshot file/lock

This results in duplicated work across processes and sustained CPU usage.

## What I observed

A local Opencode config used:

```json
"claude-context": {
  "type": "local",
  "command": [
    "nix", "shell", "github:nixos/nixpkgs/nixos-25.05#nodejs",
    "-c", "npx", "-y", "@zilliz/claude-context-mcp@latest"
  ]
}
```

With several workspaces open, I saw many `claude-context-mcp` processes active at the same time. They all shared the same snapshot state under `~/.context/mcp-codebase-snapshot.json`, so each process attempted to sync the same indexed repos.

The CPU issue seems to come from a combination of:

1. `startBackgroundSync()` being enabled unconditionally on startup
2. `handleSyncIndex()` iterating all indexed codebases from shared snapshot state
3. `reindexByChange()` / `FileSynchronizer.checkForChanges()` doing recursive full-file hashing/rescans
4. snapshot lock contention, especially because the lock retry path appears to busy-wait

### 1. Make background sync opt-in
Add a config/env flag to disable background sync entirely, for example:

- `CLAUDE_CONTEXT_BACKGROUND_SYNC=false`
- `CLAUDE_CONTEXT_SYNC_INTERVAL_MS=...`

A lot of users likely want explicit indexing/search only, not automatic periodic reindexing.

### 2. Support a real shared/remote deployment model
Right now the easiest setup is local stdio MCP per client session, which encourages N identical server instances.

It would help to support/document:

- one long-lived shared MCP server
- remote transport for multiple clients
- guidance for hosting as a singleton service instead of per-workspace process

That would avoid duplicated indexing/sync work across sessions.

### 3. Remove busy-wait lock behavior
If the snapshot lock cannot be acquired, avoid CPU-spinning while waiting. A non-busy retry or proper file lock would help a lot when multiple MCP instances are active.

### 4. Consider lighter change detection
If possible, make periodic sync cheaper than recursive read-and-hash of every tracked file in every indexed repo for every process.

### 5. Clarify which deployment model is recommended
The README could explicitly describe:

- local per-session stdio use
- shared single-host daemon use
- remote/service deployment
- tradeoffs for CPU/memory when many clients are open

## Why this matters

The current behavior is manageable with one MCP instance, but scales poorly when editor/agent tools open multiple workspaces. Even if each process uses a modest amount of CPU, the aggregate load becomes very noticeable.

Happy to help test a flag or a shared-daemon-oriented setup if that would be useful.

txhno:
Fixes #285.

This adds explicit MCP env controls for background sync so local stdio sessions do not automatically rescan indexed codebases unless the user opts in. It also makes the sync interval configurable and documents the new behavior in the MCP help text and README.

Validation:
- `pnpm install`
- `pnpm --filter @zilliz/claude-context-core build`
- `pnpm --filter @zilliz/claude-context-mcp typecheck`
- `pnpm --filter @zilliz/claude-context-mcp build`

### DIFF
--- a/packages/mcp/README.md
+++ b/packages/mcp/README.md
@@ -227,6 +227,22 @@ Notes:
 - Triggered syncs go through the same global cross-process lock as background sync, so when multiple MCP processes share `$HOME` only one process performs the work per trigger.
 - The trigger file's *contents* are ignored — only the modification event matters.
 
+#### Background Sync Configuration (Optional)
+
+By default, the MCP server runs startup + periodic background sync for compatibility with existing installations. The global cross-process sync lock ensures only one local MCP process performs a sync cycle at a time.
+
+You can tune or disable periodic polling with environment variables:
+
+```bash
+# Default: true. Set to false to disable startup + periodic polling.
+CLAUDE_CONTEXT_BACKGROUND_SYNC=false
+
+# Optional: control how often sync runs (default: 300000 = 5 minutes)
+CLAUDE_CONTEXT_SYNC_INTERVAL_MS=60000
+```
+
+For multi-instance local stdio setups, set `CLAUDE_CONTEXT_BACKGROUND_SYNC=false` and keep the trigger watcher enabled. That avoids idle polling while still allowing external tools to request immediate re-indexing by touching `~/.context/.sync-trigger`.
+
 ## Usage with MCP Clients
 
 <details>
--- a/packages/mcp/src/config.ts
+++ b/packages/mcp/src/config.ts
@@ -257,6 +257,15 @@ Environment Variables:
                           The per-codebase pathHash is preserved so multiple
                           codebases stay distinct under the same override.
 
+  MCP Sync Configuration:
+  CLAUDE_CONTEXT_BACKGROUND_SYNC
+                          Enable/disable startup + periodic background sync
+                          for indexed codebases (default: true). Set to false
+                          to disable polling while keeping trigger-based sync.
+  CLAUDE_CONTEXT_SYNC_INTERVAL_MS
+                          Background sync interval in milliseconds when enabled
+                          (default: 300000).
+
   Sync Trigger Watcher:
   CLAUDE_CONTEXT_TRIGGER_WATCHER
                           Enable/disable the ~/.context/.sync-trigger filesystem
@@ -288,5 +297,8 @@ Examples:
 
   # Start MCP server with a human-readable collection name override
   OPENAI_API_KEY=sk-xxx MILVUS_TOKEN=your-token CODE_CHUNKS_COLLECTION_NAME_OVERRIDE=my_project npx @zilliz/claude-context-mcp@latest
+
+  # Start MCP server with background sync enabled every minute
+  OPENAI_API_KEY=sk-xxx MILVUS_TOKEN=your-token CLAUDE_CONTEXT_BACKGROUND_SYNC=true CLAUDE_CONTEXT_SYNC_INTERVAL_MS=60000 npx @zilliz/claude-context-mcp@latest
         `);
 }
--- a/packages/mcp/src/sync.ts
+++ b/packages/mcp/src/sync.ts
@@ -6,8 +6,55 @@ import { SnapshotManager } from "./snapshot.js";
 import type { RequestSplitterType } from "./config.js";
 import { createRequestSplitter, resolveRequestSplitterType } from "./splitter.js";
 
+const DEFAULT_INITIAL_SYNC_DELAY_MS = 5_000;
+const DEFAULT_SYNC_INTERVAL_MS = 5 * 60 * 1000;
+const MIN_SYNC_INTERVAL_MS = 1_000;
 const DEFAULT_SYNC_LOCK_STALE_MS = 10 * 60 * 1000;
 const SYNC_LOCK_STALE_ENV = "CLAUDE_CONTEXT_SYNC_LOCK_STALE_MS";
+
+function isBackgroundSyncEnabled(): boolean {
+    const value = envManager.get("CLAUDE_CONTEXT_BACKGROUND_SYNC");
+    if (!value) {
+        return true;
+    }
+
+    switch (value.trim().toLowerCase()) {
+        case "1":
+        case "true":
+        case "yes":
+        case "on":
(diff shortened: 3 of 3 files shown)


BeamNawapat:
Thanks @txhno for tackling this — and to @jmmaloney4 for the multi-instance CPU analysis in #285 that motivated it.

I have an open PR (#332) that adds a trigger-file watcher. After reading both, I rebased it so `setupTriggerWatcher()` is invoked at the top of `startBackgroundSync()`, before the polling block. That way your `CLAUDE_CONTEXT_BACKGROUND_SYNC=false` default plays well with on-demand sync via `~/.context/.sync-trigger` — the recommended config for #285 (zero idle CPU + instant re-index when an external tool touches the trigger).

Together the two PRs give a clean opt-in matrix:

| `CLAUDE_CONTEXT_BACKGROUND_SYNC` | `CLAUDE_CONTEXT_TRIGGER_WATCHER` | Result |
|---|---|---|
| `true` | `true` (default) | polling + on-demand |
| `false` (default per #314) | `true` (default) | on-demand only — recommended for multi-instance |
| `true` | `false` | polling only |
| `false` | `false` | manual indexing only |

Happy to swap merge order either way.

zc277584121:
This is still a useful fix for #285, but the PR is currently conflicting with master and needs a rebase before it can be merged.

When rebasing, please also account for #332 if that lands first. The trigger watcher setup must happen before the `CLAUDE_CONTEXT_BACKGROUND_SYNC` gate; otherwise the new default `CLAUDE_CONTEXT_BACKGROUND_SYNC=false` would return early and disable on-demand trigger sync as well.

The intended combined flow should be:

```ts
this.setupTriggerWatcher();

if (!isBackgroundSyncEnabled()) {
    return;
}

// initial sync + periodic interval only when enabled
```

I also verified this branch by itself with:

```
pnpm install --frozen-lockfile
pnpm --filter @zilliz/claude-context-mcp build
```

That build passes, so the blocker is the rebase/combined behavior rather than the basic TypeScript build.

zc277584121:
Thanks again for working on this.

We shipped v0.1.10 with a lighter compatibility-oriented fix for the multi-process background sync problem: a global cross-process sync lock. Background sync remains enabled for existing users, but only one local MCP server process should perform background sync at a time; other processes skip the cycle while the lock is held.

This should address a large part of the CPU churn and contention described in #285 without changing the default background sync behavior. Please try the latest version and let us know if you still see cases where making background sync opt-in is needed.

txhno:
Rebased this PR onto current `master` (now includes the v0.1.10 global cross-process sync lock) and resolved the conflicts.

I kept the upstream lock behavior intact, while preserving this PR's opt-in background sync behavior via `CLAUDE_CONTEXT_BACKGROUND_SYNC=true` and `CLAUDE_CONTEXT_SYNC_INTERVAL_MS`.

Validation run locally:
- `pnpm --filter @zilliz/claude-context-mcp build`
- default startup returns before scheduling background sync
- `CLAUDE_CONTEXT_BACKGROUND_SYNC=true CLAUDE_CONTEXT_SYNC_INTERVAL_MS=1000` schedules periodic sync
- concurrent `handleSyncIndex()` calls still use the global lock; the second call skips while the first sync runs

For #332 compatibility, this branch leaves the background-sync gate in `startBackgroundSync()`. If #332 lands first or is merged together, `setupTriggerWatcher()` should stay before the `isBackgroundSyncEnabled()` return, as discussed above.

zc277584121:
Thanks for the rebase and for validating this on top of the global sync lock.

#332 has now been merged, so this PR is conflicting again in `packages/mcp/src/sync.ts`, `packages/mcp/README.md`, and `packages/mcp/src/config.ts`.

When rebasing, please preserve the trigger watcher setup from #332 before any polling/background-sync gate:

```ts
this.setupTriggerWatcher();

if (!isBackgroundSyncEnabled()) {
    return;
}

// startup sync + periodic sync
```

That keeps on-demand trigger sync available even when periodic polling is disabled.

One compatibility concern remains: changing `CLAUDE_CONTEXT_BACKGROUND_SYNC` to default off would change existing behavior. Today users get startup + periodic background sync without extra configuration. If this PR makes it opt-in by default, users who do not configure the trigger watcher hook may stop getting automatic index refreshes.

A more compatible version would still add `CLAUDE_CONTEXT_BACKGROUND_SYNC` and `CLAUDE_CONTEXT_SYNC_INTERVAL_MS`, but keep background sync enabled by default for now. Multi-instance users can then explicitly set `CLAUDE_CONTEXT_BACKGROUND_SYNC=false` and rely on the trigger watcher from #332 plus the global sync lock.

So my recommendation is:

- rebase on current master after #332
- keep `setupTriggerWatcher()` before the polling gate
- consider making the default for `CLAUDE_CONTEXT_BACKGROUND_SYNC` remain enabled, and document disabling it as the recommended multi-instance optimization

With those changes, this becomes a safer configuration improvement instead of a breaking default-behavior change.

txhno:
Updated this PR against current `master` and resolved the conflicts in `packages/mcp/src/sync.ts`, `packages/mcp/src/config.ts`, and `packages/mcp/README.md`.

Changes made:

- preserved `setupTriggerWatcher()` before the background-sync gate, so trigger-based sync stays available when polling is disabled
- kept background sync enabled by default for compatibility with existing users
- kept `CLAUDE_CONTEXT_BACKGROUND_SYNC=false` as the opt-out path for multi-instance setups that want trigger-only sync
- documented the background sync controls alongside the trigger watcher docs

Validation:

- `pnpm --filter @zilliz/claude-context-mcp build`
- `git diff --check upstream/master..HEAD`

PR #318, opened 2026-04-22
fix(mcp): drop unreachable throw in setTimeout-wrapped initial sync

### LINKED ISSUE #256: Error thrown inside setTimeout goes nowhere
## Problem

In `startBackgroundSync`, error is re-thrown inside setTimeout callback.

## Current Behavior

```js
// sync.ts:119-131
setTimeout(async () => {
    try {
        await this.handleSyncIndex();
    } catch (error) {
        if (/* unexpected error */) {
            throw error; // Goes nowhere!
        }
    }
}, 5000);
```

Throwing inside async setTimeout callback doesn't propagate - error is silently lost.

## Impact

Unexpected errors during initial sync are swallowed, making debugging difficult.

## Expected Behavior

Log error and/or emit event instead of throwing.

## Location

- `packages/mcp/src/sync.ts:128-129`

mvanhorn:
## Summary

Drops a `throw error` inside the initial-sync `setTimeout` callback in `packages/mcp/src/sync.ts`. The callback runs with no caller, so the throw has nowhere to propagate. The preceding `console.error(...)` is the terminal signal for unexpected failures.

## Why this matters

Reporter in [#256](https://github.com/zilliztech/claude-context/issues/256) called out the bug with an exact file:line pointer. Unexpected errors during the 5s-deferred initial sync are silently lost today, not surfaced. This is a pure log-and-continue fix.

## Changes

```diff
                 } else {
                     console.error('[SYNC-DEBUG] Initial sync failed with unexpected error:', error);
-                    throw error;
+                    // Do not re-throw here: this callback runs via setTimeout with no caller to propagate to.
                 }
```

No other changes. No behavior regression: the throw was already unreachable.

## Testing

`pnpm install --frozen-lockfile && pnpm build` passes locally on Node 22.

## Related sync.ts PRs

Aware of #314, #274, #268, #234 also touching `packages/mcp/src/sync.ts`. None of them address the setTimeout throw (#314 preserves it at the relocated line). Whichever lands first, the other rebases cleanly.

Fixes #256

### DIFF
--- a/packages/mcp/src/sync.ts
+++ b/packages/mcp/src/sync.ts
@@ -126,7 +126,7 @@ export class SyncManager {
                     console.log('[SYNC-DEBUG] Collection not yet established, this is expected for new cluster users. Will retry on next sync cycle.');
                 } else {
                     console.error('[SYNC-DEBUG] Initial sync failed with unexpected error:', error);
-                    throw error;
+                    // Do not re-throw here: this callback runs via setTimeout with no caller to propagate to.
                 }
             }
         }, 5000); // Initial sync after 5 seconds


mvanhorn:
Appreciate getting the dead-code path out of the initial sync.

PR #319, opened 2026-04-22
fix(core): tolerate gRPC timeout in checkCollectionLimit and clean up orphans

### LINKED ISSUE #289: DEADLINE_EXCEEDED on checkCollectionLimit() - gRPC timeout on collection creation (Free tier, macOS)
## Bug Description

Consistent `DEADLINE_EXCEEDED` error when attempting to index a codebase. The error occurs during the pre-flight `checkCollectionLimit()` call before any actual indexing begins. A different codebase on the same cluster was successfully indexed 4 days prior.

## Environment

- **OS**: macOS Darwin 25.3.0 (Apple Silicon)
- **Node.js**: v24.3.0
- **Package**: `@zilliz/claude-context-mcp@0.1.6`
- **Zilliz Cloud**: Free tier, `gcp-us-west1`, Milvus 2.6.x compatible
- **Cluster**: `in03-e63f077fc4f1b69`

## Cluster State

- **Collections**: 1 of 5 allowed (`hybrid_code_chunks_9429127a`, status: Loaded)
- **Storage**: 0.01 GB of 5 GB
- **Read vCUs**: ~12 (minimal load)
- No resource limits hit

## Error

```
Error validating collection creation: 4 DEADLINE_EXCEEDED: Deadline exceeded after 15.001s,remote_addr=34.111.198.99:443
```

## Root Cause Trace

The error originates in `milvus-vectordb.js:607` (`checkCollectionLimit()`), which creates a dummy collection as a pre-flight limit check:

```javascript
async checkCollectionLimit() {
    const collectionName = `dummy_collection_${Date.now()}`;
    // creates a trivial 2-field collection, then drops it
    await this.client.createCollection(createCollectionParams);  // <-- times out here
}
```

This is called from `handlers.js:249` before any real indexing starts:

```javascript
const canCreateCollection = await this.context.getVectorDatabase().checkCollectionLimit();
```

The `MilvusClient` at `milvus-vectordb.js:20` is created with no `timeout` parameter, so the 15s gRPC deadline is the SDK default.

## Reproduction Steps

1. Have an existing indexed codebase (collection loaded and working)
2. Attempt to index a second, different codebase: `index_codebase({ path: "/path/to/project" })`
3. Consistently fails with `DEADLINE_EXCEEDED` on the `checkCollectionLimit()` dummy collection creation
4. Tried with both `ast` and `langchain` splitters, with and without `force: true`, with and without `ignorePatterns` — all produce the same error
5. Indexing a subdirectory also fails identically

## Key Observation

- Another codebase (`~/gaylonphotos`, 297 files, 1241 chunks) was successfully indexed on this same cluster on Mar 22, 2026
- The failing codebase (`~/BTC-dashboard`) was previously indexed (196k chunks, Feb 20) but was cleared during a force-reindex attempt
- Since then, every attempt to create any new collection times out
- `get_indexing_status` confirms the codebase is not indexed
- `clear_index` confirms nothing to clear

## Related

- Issue #154 — same `DEADLINE_EXCEEDED` error, also on Node v24.x
- Issue #126 — similar collection creation failure (different error format)

## Suggested Improvements

1. **Configurable gRPC timeout** — allow setting a timeout > 15s via env var (e.g., `MILVUS_TIMEOUT_MS`)
2. **Skip dummy collection check** — `checkCollectionLimit()` could use `listCollections()` + count instead of create/drop, avoiding the gRPC-heavy collection creation path
3. **REST API fallback** — `MilvusRestfulVectorDatabase` exists in `claude-context-core` but isn't wired up as a fallback when gRPC fails

### LINKED ISSUE #290: DEADLINE_EXCEEDED error when indexing codebase - gRPC timeout during collection creation
## Bug Description

When trying to index a codebase using the `index_codebase` tool, it consistently fails with a gRPC `DEADLINE_EXCEEDED` error during collection creation validation.

## Error Message

```
Error validating collection creation: 4 DEADLINE_EXCEEDED: Deadline exceeded after 15.001s, metadata filters: 0.001s, remote_addr=34.111.198.99:443
```

## Environment

- **MCP Package**: `@zilliz/claude-context-mcp@latest`
- **Embedding Provider**: VoyageAI (`voyage-code-3`)
- **Milvus Address**: Zilliz serverless cluster (`*.serverless.gcp-us-west1.cloud.zilliz.com`)
- **OS**: macOS Darwin 24.6.0
- **Claude Code**: CLI

## Steps to Reproduce

1. Configure `claude-context` MCP with a Zilliz serverless cluster
2. Call `index_codebase` with an absolute path to a local directory
3. Error is thrown immediately after 15 seconds

## Network Diagnostics

The Zilliz cluster endpoint is reachable:
- ✅ DNS resolves correctly
- ✅ TCP connection to port 443 succeeds
- ✅ TLS handshake completes successfully (TLSv1.3)
- ✅ Ping responds with ~22ms latency, 0% packet loss

The issue appears to be at the **gRPC layer** — the HTTP/HTTPS connection works fine, but the gRPC call times out during `collection creation validation`.

## Expected Behavior

The codebase should be indexed successfully.

## Actual Behavior

The tool times out after 15 seconds with `DEADLINE_EXCEEDED` on every attempt.

## Possible Cause

The default gRPC deadline of 15 seconds may be too short, especially when:
- The Zilliz serverless cluster is warming up / resuming from a paused state
- There is high latency on the gRPC metadata/collection API

## Suggested Fix

- Increase the gRPC deadline timeout for collection creation validation (e.g., 30–60 seconds)
- Add retry logic with exponential backoff for transient timeout errors
- Provide a clearer error message indicating the cluster may be paused/warming up

mvanhorn:
## Summary

Makes `checkCollectionLimit()` tolerant of slow or congested Zilliz clusters. The preflight dummy-create uses the Milvus client's default 15s gRPC deadline; on free-tier or serverless clusters that deadline fires before the RPC can settle and the user sees a `DEADLINE_EXCEEDED` before any indexing starts.

## Why this matters

Two open issues with the same symptom, same Zilliz cluster address (`34.111.198.99:443`), same `checkCollectionLimit()` call:
- [#290](https://github.com/zilliztech/claude-context/issues/290) - Zilliz serverless, VoyageAI, macOS
- [#289](https://github.com/zilliztech/claude-context/issues/289) - Zilliz Free tier, same cluster, same 15s deadline

@aud1ence did the root-cause trace on #290 and located it in `packages/core/src/vectordb/milvus-vectordb.ts`.

## Changes

`checkCollectionLimit()` now wraps the dummy `createCollection` in a `Promise.race` with a configurable deadline:

- New env var `MILVUS_COLLECTION_LIMIT_CHECK_TIMEOUT_MS`, default `15000` (matches current behavior for anyone not setting it)
- On timeout or DEADLINE_EXCEEDED, warn and return `true`. The subsequent real `createCollection` surfaces a cleaner quota error if the user is actually over the limit
- Quota detection path (`exceeded the limit number of collections`) unchanged
- Other errors still re-thrown
- Orphan safety: if the gRPC create eventually resolves after the race timed out, a detached best-effort handler drops the late-arriving `dummy_collection_*`. The handler is gated by a `timedOut` flag so it only runs on the timeout path and never races the main-path drop on fast successes

Mirrors the REST backend's existing unimplemented-stub behavior (returns `true` with a warning).

## Testing

`pnpm install --frozen-lockfile && pnpm build` passes locally on Node 22.

Manual repro of the happy path (`checkCollectionLimit` completes under deadline, dummy created and dropped, return `true`) and the timeout path (set `MILVUS_COLLECTION_LIMIT_CHECK_TIMEOUT_MS=1`, expect warn-and-return-`true`) was reasoned through, not executed end-to-end against a live cluster.

## Docs

- `docs/getting-started/environment-variables.md`: added row for the new env var
- `packages/mcp/README.md`: added an example entry

## Preempting a couple of reviewer questions

- Why not swap to `listCollections().length` as @aud1ence suggested? Different semantics - Milvus collection limits count aliased/loading collections in ways that a naive length check would miss. That's a larger rework and needs its own design discussion.
- What if the user really is over quota when we time out? The next real `createCollection` for indexing fires anyway and returns the actual `exceeded the limit number of collections` error with the proper message. Trading one obscure timeout error for a clearer quota error.

Fixes #289
Fixes #290

### DIFF
--- a/docs/getting-started/environment-variables.md
+++ b/docs/getting-started/environment-variables.md
@@ -47,6 +47,7 @@ Claude Context supports a global configuration file at `~/.context/.env` to simp
 |----------|-------------|---------|
 | `MILVUS_TOKEN` | Milvus authentication token. Get [Zilliz Personal API Key](https://github.com/zilliztech/claude-context/blob/master/assets/signup_and_get_apikey.png) | Recommended |
 | `MILVUS_ADDRESS` | Milvus server address. Optional when using Zilliz Personal API Key | Auto-resolved from token |
+| `MILVUS_COLLECTION_LIMIT_CHECK_TIMEOUT_MS` | Timeout for gRPC pre-check in `checkCollectionLimit()` before indexing begins | `15000` |
 
 ### Ollama (Optional)
 | Variable | Description | Default |
--- a/packages/core/src/vectordb/milvus-vectordb.ts
+++ b/packages/core/src/vectordb/milvus-vectordb.ts
@@ -742,6 +742,8 @@ export class MilvusVectorDatabase implements VectorDatabase {
             throw new Error('MilvusClient is not initialized. Call ensureInitialized() first.');
         }
 
+        const parsedTimeoutMs = Number(process.env.MILVUS_COLLECTION_LIMIT_CHECK_TIMEOUT_MS);
+        const timeoutMs = Number.isFinite(parsedTimeoutMs) && parsedTimeoutMs > 0 ? parsedTimeoutMs : 15000;
         const collectionName = `dummy_collection_${Date.now()}`;
         const createCollectionParams = {
             collection_name: collectionName,
@@ -761,8 +763,39 @@ export class MilvusVectorDatabase implements VectorDatabase {
             ]
         };
 
+        let timeoutHandle: ReturnType<typeof setTimeout> | undefined;
+        let timedOut = false;
         try {
-            await this.client.createCollection(createCollectionParams);
+            const createPromise = this.client.createCollection(createCollectionParams);
+            // Best-effort late cleanup ONLY if the timeout fires first. Gated by `timedOut`
+            // so we do not race the normal-path drop below on fast successes.
+            void createPromise
+                .then(async () => {
+                    if (!timedOut) return;
+                    try {
+                        if (!this.client) return;
+                        if (await this.client.hasCollection({ collection_name: collectionName })) {
+                            await this.client.dropCollection({
+                                collection_name: collectionName,
+                            });
+                        }
+                    } catch {
+                        // Best effort only: orphan cleanup is not user-visible.
+                    }
+                })
+                .catch(() => {
+                    // createCollection failed; nothing to clean up.
+                });
+
+            await Promise.race([
+                createPromise,
+                new Promise<never>((_, reject) => {
+                    timeoutHandle = setTimeout(() => {
+                        timedOut = true;
+                        reject(new Error(`checkCollectionLimit timeout after ${timeoutMs}ms`));
+                    }, timeoutMs);
+                }),
+            ]);
             // Immediately drop the collection after successful creation
             if (await this.client.hasCollection({ collection_name: collectionName })) {
                 await this.client.dropCollection({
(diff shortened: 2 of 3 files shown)


mvanhorn:
Thanks again. The orphan cleanup on gRPC timeout was overdue.

Issue #271, opened 2026-02-18
Support shared indexes across machines via git-remote-based collection naming


RyanColton:
## Problem

Collection names are currently derived from the MD5 hash of the absolute local file path:

```js
// context.js
const normalizedPath = path.resolve(codebasePath);
const hash = crypto.createHash('md5').update(normalizedPath).digest('hex');
return `${prefix}_${hash.substring(0, 8)}`;
```

This means two users indexing the same codebase on different machines (e.g. `/Users/alice/monorepo` vs `/Users/bob/projects/monorepo`) produce different collection names and cannot share indexes. On a team, this results in:

- Redundant collections consuming Zilliz Cloud quota (especially painful on the free tier's 5-collection limit)
- Redundant embedding API calls (and cost) for the same codebase
- No way for a team to maintain a single shared index

## Proposed Solution

Use a **stable, machine-independent identifier** for collection naming when available. The most natural candidate for code repositories is the **git remote URL**:

```js
// Example: derive hash from git remote origin URL instead of local path
const remoteUrl = execSync('git -C <path> remote get-url origin').toString().trim();
const hash = crypto.createHash('md5').update(remoteUrl).digest('hex');
```

This could be implemented as a fallback chain:
1. If the user provides an explicit `collectionName` parameter — use that
2. If the indexed path is a git repo — hash the remote origin URL (+ optional subpath for monorepos)
3. Fall back to the current behavior (hash of absolute path)

For monorepos, the hash input could be `${remoteUrl}:${relativeSubpath}` so that indexing different subdirectories still produces distinct collections.

## Additional Benefits

- Teams share a single Zilliz Cloud token and automatically reuse each other's indexes
- Re-cloning or moving a repo doesn't orphan the old collection
- Git worktrees of the same repo would share indexes instead of consuming extra slots

## Related Bug: `[object Object]` error when collection limit is reached

While debugging this, we discovered that when the Zilliz free tier collection limit (5) is hit, the error surfaces as:

```
Error validating collection creation: [object Object]
```

This is because `checkCollectionLimit()` in `milvus-vectordb.js` checks `error.message` for the "exceeded the limit" pattern, but the Milvus SDK returns a plain object (not an `Error` instance) with the info in `.reason` and `.detail` instead of `.message`:

```json
{
  "error_code": "UnexpectedError",
  "reason": "exceeded the limit number of collections[dbName=...][limit=5]",
  "code": 102,
  "retriable": false,
[...]

PR #320, opened 2026-04-22
feat(core): support custom collection-name override via env var

### LINKED ISSUE #299: Feature: Allow custom collection name (use folder name instead of hash)
## Problem

Currently, `getCollectionName()` in `context.js` generates collection names as `code_chunks_<md5-hash>` based on the absolute path:

```js
getCollectionName(codebasePath) {
    const normalizedPath = path.resolve(codebasePath);
    const hash = crypto.createHash('md5').update(normalizedPath).digest('hex');
    const prefix = isHybrid ? 'hybrid_code_chunks' : 'code_chunks';
    return `${prefix}_${hash.substring(0, 8)}`;
}
```

This makes it impossible to identify which collection belongs to which project in the Zilliz Cloud console. For example, `code_chunks_a1b2c3d4` gives no indication that it's the `sda-service` project.

## Proposed Solution

Allow users to configure a human-readable collection name, falling back to the current hash-based naming. Options:

1. **Use the folder name by default**: `code_chunks_sda_service` instead of `code_chunks_a1b2c3d4`
2. **Add an env var** like `COLLECTION_NAME_STRATEGY=folder|hash` (default: `folder`)
3. **Add a parameter** to `index_codebase` tool: `collectionName` (optional)

Using the folder basename (`path.basename(codebasePath)`) would make collection management much more intuitive in the Zilliz Cloud console, especially when managing multiple indexed codebases.

## Environment

- `@zilliz/claude-context-mcp` via npx
- Zilliz Cloud Serverless (AWS eu-central-1)
- Gemini embeddings

mvanhorn:
## Summary

Adds an opt-in override for the collection name so the Zilliz/Milvus console shows `code_chunks_my_project_a1b2c3d4` instead of just `code_chunks_a1b2c3d4`. Default behavior (no override set) is byte-identical to today.

## Why this matters

Today `getCollectionName()` in `packages/core/src/context.ts` returns `code_chunks_<md5(abs_path).slice(0,8)>`. In [#299](https://github.com/zilliztech/claude-context/issues/299) the reporter points out this makes the Zilliz console unusable for humans - `code_chunks_a1b2c3d4` tells you nothing about which project it belongs to. Neighbor issue [#271](https://github.com/zilliztech/claude-context/issues/271) asks for the same readability from the team-sharing angle (git-remote-based naming).

This PR is a minimal opt-in readability win. It intentionally does NOT try to solve #271's cross-machine sharing story - that's a separate design.

## Changes

New resolution order in `getCollectionName()`:

1. `ContextConfig.collectionNameOverride` (explicit constructor arg)
2. `CODE_CHUNKS_COLLECTION_NAME_OVERRIDE` env var
3. Plain md5 path hash (today's default - unchanged)

The override:

- Gets the existing `code_chunks_` / `hybrid_code_chunks_` prefix so it can't collide with auto-named collections
- Is sanitized per Milvus naming rules (alphanumeric + underscore, truncated to fit the 255-char name limit)
- Keeps the per-codebase path hash appended: final shape is `code_chunks_<override>_<pathHash>`. An MCP server that indexes multiple repos under the same override won't collapse them onto one collection
- Deduplicates sanitization warnings per (source, input) to avoid log spam

Wires the env var through the MCP config layer (`packages/mcp/src/config.ts` + `index.ts`) and adds it to the help text, env-var docs, and README examples.

## Testing

`pnpm install --frozen-lockfile && pnpm build` passes locally on Node 22. No Milvus integration harness exists in the repo - the new sanitizer has no test coverage yet; happy to add a small unit test if the repo adds a test framework for the core package.

## Explicit non-goals

- No auto-derive-from-folder-name behavior. Two folders named `monorepo` on two machines would clash - opt-in prevents accidents
- No git-remote-based naming (that's #271, larger scope, needs its own design discussion)
- No migration of existing hash-named collections. Setting or unsetting the override creates a fresh collection at the new name - documented as "opt-in-sticky" in the README

## Preempting reviewer questions

- **What happens on reconnect if I unset the env var?** The next `getCollectionName()` call returns to plain `code_chunks_<pathHash>` for that codebase path. The override-named collection data stays in Milvus but the MCP server no longer references it. Documented in the README.
- **Why keep the path hash suffix when the whole point is readability?** Without it, multi-codebase MCP sessions silently merge data across codebases. The hash costs 9 characters and keeps the override safe. The name still reads as "this is the my_project collection for path-variant a1b2c3d4" in the Zilliz console.
- **Why not solve #271 at the same time?** Different problem shape (cross-machine git-remote lookups, config format for multi-project teams). Layering that on top of this opt-in is trivial once the maintainer decides the approach.

Fixes #299

### DIFF
--- a/docs/getting-started/environment-variables.md
+++ b/docs/getting-started/environment-variables.md
@@ -63,6 +63,9 @@ Claude Context supports a global configuration file at `~/.context/.env` to simp
 | `SPLITTER_TYPE` | Code splitter type: `ast`, `langchain` | `ast` |
 | `CUSTOM_EXTENSIONS` | Additional file extensions to include (comma-separated, e.g., `.vue,.svelte,.astro`) | None |
 | `CUSTOM_IGNORE_PATTERNS` | Additional ignore patterns (comma-separated, e.g., `temp/**,*.backup,private/**`) | None |
+| `CODE_CHUNKS_COLLECTION_NAME_OVERRIDE` | Optional custom prefix for collection names. Produces `code_chunks_<suffix>_<pathHash>` or `hybrid_code_chunks_<suffix>_<pathHash>` after sanitization. The path hash stays appended so collections remain unique per codebase even when the override is set | None |
+
+When `CODE_CHUNKS_COLLECTION_NAME_OVERRIDE` is set, Claude Context writes to an override-named collection instead of the default `code_chunks_<pathHash>`. The per-codebase `<pathHash>` suffix is preserved to keep multiple codebases distinct under the same override. If you later unset the variable, Claude Context returns to the plain hash-based naming for that path.
 
 ## 🚀 Quick Setup
 
--- a/packages/core/src/context.ts
+++ b/packages/core/src/context.ts
@@ -94,14 +94,19 @@ export interface ContextConfig {
     ignorePatterns?: string[];
     customExtensions?: string[]; // New: custom extensions from MCP
     customIgnorePatterns?: string[]; // New: custom ignore patterns from MCP
+    collectionNameOverride?: string; // Optional: custom collection name suffix
 }
 
 export class Context {
+    private static readonly MAX_COLLECTION_NAME_LENGTH = 255;
+
     private embedding: Embedding;
     private vectorDatabase: VectorDatabase;
     private codeSplitter: Splitter;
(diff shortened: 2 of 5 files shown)


mvanhorn:
Thanks for landing the env override for collection-name.

PR #325, opened 2026-04-23
feat: add maxDepth option to limit codebase traversal depth

### LINKED ISSUE #324: Add --max-depth flag to limit codebase traversal depth
## Problem\n\nClaude Context indexes the entire codebase without a way to limit traversal depth. For large monorepos, this can be slow and include files that are irrelevant to the current task.\n\n## Solution\n\nAdd a `--max-depth` CLI flag (default: unlimited) that stops traversal at the specified directory depth. This allows users to scope the context to the relevant subtree.\n\n## Use case\n\nA developer working in `packages/api/src/` doesn't need the entire monorepo context — `--max-depth 4` would capture the relevant package scope without pulling in unrelated `packages/web/` or `packages/shared/` content.\n\n## Notes\n\nThis would be a non-breaking addition; existing behavior remains unchanged when the flag is omitted.

c6zks4gssn-droid:
Fixes #324. Adds maxDepth to ContextConfig and MAX_DEPTH env var to MCP server. Non-breaking: existing behavior unchanged when option is omitted.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,2477 @@
+{
+    "name": "claude-context",
+    "version": "0.1.7",
+    "lockfileVersion": 3,
+    "requires": true,
+    "packages": {
+        "": {
+            "name": "claude-context",
+            "version": "0.1.7",
+            "license": "MIT",
+            "devDependencies": {
+                "@types/node": "^20.0.0",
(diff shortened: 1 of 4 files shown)


zc277584121:
Please remove package-lock.json from this PR — this project uses pnpm, not npm. Happy to review the actual code changes once cleaned up.

PR #327, opened 2026-04-24
fix(mcp): sync from cloud in handleGetIndexingStatus

### LINKED ISSUE #252: Inconsistent cloud sync - handleGetIndexingStatus skips sync
## Problem

`handleSearchCode` and `handleIndexCodebase` call `syncIndexedCodebasesFromCloud()` before operations, but `handleGetIndexingStatus` does not.

## Current Behavior

```js
// handlers.ts:144 - handleIndexCodebase
await this.syncIndexedCodebasesFromCloud();

// handlers.ts:414 - handleSearchCode  
await this.syncIndexedCodebasesFromCloud();

// handlers.ts:691 - handleGetIndexingStatus
// NO sync call
```

## Impact

Status check may return stale/incorrect data compared to actual index/search operations.

## Expected Behavior

Either:
1. All handlers sync before operation (consistent but slower)
2. No handlers sync inline (rely on background sync only)

## Location

- `packages/mcp/src/handlers.ts:691-799`

mvanhorn:
## Summary

`handleGetIndexingStatus` reads local snapshot state without first calling `syncIndexedCodebasesFromCloud()`, so a freshly-started MCP server can return stale or empty indexing status even when cloud collections exist.

## Changes

Added a single `await this.syncIndexedCodebasesFromCloud();` at the top of `handleGetIndexingStatus`'s try block, after path validation and before snapshot reads. Mirrors the pattern already in place at `packages/mcp/src/handlers.ts:316` (`handleIndexCodebase`) and `:605` (`handleSearchCode`). The existing try/catch keeps status reporting resilient if the sync call fails.

```ts
// packages/mcp/src/handlers.ts, inside handleGetIndexingStatus
+ await this.syncIndexedCodebasesFromCloud();

  const status = this.snapshotManager.getCodebaseStatus(absolutePath);
  const info   = this.snapshotManager.getCodebaseInfo(absolutePath);
```

## Testing

`pnpm build` passes.

Manual verification:
1. Start MCP server against a cloud workspace that already has an indexed collection
2. Before this change: `get_indexing_status` returns empty until another handler (search or index) triggers the cloud sync
3. With this change: `get_indexing_status` returns reconciled state on the first call

Fixes #252

### DIFF
--- a/packages/mcp/src/handlers.ts
+++ b/packages/mcp/src/handlers.ts
@@ -938,6 +938,8 @@ export class ToolHandlers {
                 };
             }
 
+            await this.syncIndexedCodebasesFromCloud();
+
             // Check indexing status using new status system
             const status = this.snapshotManager.getCodebaseStatus(absolutePath);
             const info = this.snapshotManager.getCodebaseInfo(absolutePath);


mvanhorn:
Thanks for the merge!

PR #331, opened 2026-04-25
feat: add disk-based embedding cache

BeamNawapat:
## Summary

Add a transparent, disk-based embedding cache to skip redundant embedding API calls when re-indexing the same code. On a re-index, only chunks whose content has not been embedded before hit the API; cached chunks load from disk in milliseconds.

## Motivation

Embedding API calls are the slowest and most expensive step in indexing. When a codebase is re-indexed (after `force: true`, switching machines, or recovering from a failed run), every chunk is sent through the embedding provider again — even if its content is byte-identical to a previous run.

In practice this wastes:
- **API quota / cost** — large monorepos can re-burn $1–$5 per re-index
- **Latency** — re-indexing a 10k-file repo against VoyageAI takes minutes purely for embeddings
- **Provider rate limits** — easy to hit on Gemini / OpenAI free tiers

A small content-addressed cache keyed by `SHA256(content)` per `(provider, dimension)` eliminates this waste entirely for unchanged chunks.

## Changes

- **`packages/core/src/embedding/embedding-cache.ts`** *(new, ~130 LOC)* — `EmbeddingCache` class with `get`, `set`, `getBatch`, `cleanup` methods. Storage: `~/.context/embedding-cache/{provider}_{dimension}/XX/{sha256}.json` (hierarchical to avoid single-dir overflow). No external dependencies — uses Node `fs` / `crypto` / `path` / `os`.
- **`packages/core/src/embedding/index.ts`** — Export `EmbeddingCache`.
- **`packages/core/src/context.ts`** — Initialize `EmbeddingCache` in constructor keyed by `${provider}_${dimension}`. New private `cachedEmbedBatch()` wraps `embedding.embedBatch()`: returns cached vectors instantly, only sends uncached chunks to the API. Indexing path now calls `cachedEmbedBatch()` instead of `embedding.embedBatch()` directly. Async TTL cleanup runs once on startup (non-blocking).

Behavior:
- Per-model isolation prevents cross-contamination when switching providers (e.g., `voyage-code-3` vs `text-embedding-3-small`).
- Best-effort design: any cache I/O error falls back to a normal API call. Corrupted JSON, missing files, permission errors all degrade gracefully.
- Hit rate logged per batch: `[Cache] 75% hit (3/4 cached, 1 embedded)`.
- Stale entries auto-removed on startup based on `EMBEDDING_CACHE_MAX_AGE_DAYS` (default 30).

## Configuration

| Env var | Default | Purpose |
|---|---|---|
| `EMBEDDING_CACHE` | `true` | Enable/disable. Set to `false` to opt out completely. |
| `EMBEDDING_CACHE_DIR` | `~/.context/embedding-cache` | Storage location. |
| `EMBEDDING_CACHE_MAX_AGE_DAYS` | `30` | TTL for cleanup-on-startup. Set `0` to disable cleanup. |

## Usage

Zero-config — cache is on by default. Re-indexing the same content shows the hit rate:

```
[Context] 💾 Embedding cache enabled for model: VoyageAI_1024
[Cache] ✅ All 47 embeddings from cache
[Cache] 88% hit (44/50 cached, 6 embedded)
```

Disable temporarily:
```bash
EMBEDDING_CACHE=false npx @zilliz/claude-context-mcp@latest
```

Move cache to a shared location:
```bash
EMBEDDING_CACHE_DIR=/mnt/team-cache/embeddings npx @zilliz/claude-context-mcp@latest
```

## Test plan

- [x] `pnpm build` passes (core + mcp)
- [ ] Index a small repo, then re-index → cache hit rate should be ~100%
- [ ] Edit one file, re-index → only changed chunks re-embedded
- [ ] Switch `EMBEDDING_PROVIDER` → new cache directory created, old one untouched
- [ ] `EMBEDDING_CACHE=false` → no cache directory created, no `[Cache]` logs
- [ ] Delete `~/.context/embedding-cache/` mid-run → next batch falls back to API gracefully

## Notes for reviewers

- Cache files are JSON `{"v": [vector...], "d": dimension}` — small (~6KB per 1024-dim float vector) but consider compression in a follow-up if storage becomes a concern.
- No write locking; concurrent indexers writing the same key would race, but the result is functionally identical so this is intentionally not guarded.
- Old cache entries from previous models are *not* deleted on provider switch (only TTL cleanup applies). Trade-off: simpler logic vs slightly larger disk usage. A `clear_cache` MCP tool could be added in a follow-up.

### DIFF
--- a/packages/core/src/context.ts
+++ b/packages/core/src/context.ts
@@ -166,6 +166,7 @@ export class Context {
         this.embeddingCache = new EmbeddingCache(cacheModel);
         if (this.embeddingCache.isEnabled()) {
             console.log(`[Context] 💾 Embedding cache enabled for model: ${cacheModel}`);
+            this.embeddingCache.cleanup().catch(() => {});
         }
     }
 
--- a/packages/core/src/context.ts
+++ b/packages/core/src/context.ts
@@ -162,8 +162,19 @@ export class Context {
         }
 
         // Initialize embedding cache
-        const cacheModel = `${this.embedding.getProvider()}_${this.embedding.getDimension()}`;
-        this.embeddingCache = new EmbeddingCache(cacheModel);
+        this.initEmbeddingCache();
+    }
+
+    /**
+     * (Re)create the embedding cache keyed by current provider + dimension.
+     * Called on construction and again whenever the embedding instance changes
+     * via updateEmbedding(), so cached vectors from a previous model never bleed
+     * into the new one.
+     */
+    private initEmbeddingCache(): void {
+        const dimension = this.embedding.getDimension();
+        const cacheModel = `${this.embedding.getProvider()}_${dimension}`;
+        this.embeddingCache = new EmbeddingCache(cacheModel, undefined, dimension);
         if (this.embeddingCache.isEnabled()) {
             console.log(`[Context] 💾 Embedding cache enabled for model: ${cacheModel}`);
             this.embeddingCache.cleanup().catch(() => {});
@@ -581,6 +592,10 @@ export class Context {
 
     /**
      * Embed batch with disk cache. Only calls API for uncached chunks.
+     * Also dedupes duplicate strings within the same batch — without this, two
+     * identical chunks in one batch would each hit the API since neither is
+     * cached at the start of the call. Common in monorepos with re-exports or
+     * generated boilerplate.
      */
     private async cachedEmbedBatch(contents: string[]): Promise<EmbeddingVector[]> {
         if (!this.embeddingCache || !this.embeddingCache.isEnabled()) {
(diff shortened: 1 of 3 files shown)


BeamNawapat:
Addressed Copilot review feedback in ea99ede:

- Full SHA256 in filename (no truncation/collision risk)
- get() validates JSON shape (Array.isArray, dimension match)
- cleanup() rewritten with fs.promises (truly async, no event-loop block) and guards maxAgeDays <= 0
- updateEmbedding() reinitializes the cache so model switches can't serve stale vectors
- cachedEmbedBatch() dedupes duplicate strings within a single batch so identical chunks share one API call

All checks still green.

PR #333, opened 2026-04-25
feat: deduplicate overlapping search results

BeamNawapat:
## Summary

Deduplicate overlapping chunks in semantic and hybrid search results. When two results from the same file overlap by more than 50% of their line range, keep only the higher-scored one.

## Motivation

The AST splitter and the langchain splitter both produce overlapping chunks by design (300-line overlap window in the default `AstCodeSplitter(2500, 300)`). When a query strongly matches a region near a chunk boundary, the search routinely returns 2–3 nearly-identical chunks of the same code — wasting LLM context and pushing more relevant unique results out of the top-N.

Example before this change (top 5 results for `"qdrant collection setup"`):
```
1. score=0.91  qdrant-vectordb.ts  L120-180
2. score=0.89  qdrant-vectordb.ts  L150-210   ← overlaps result 1
3. score=0.87  qdrant-vectordb.ts  L180-240   ← overlaps result 1
4. score=0.85  qdrant-vectordb.ts  L210-270   ← overlaps result 2
5. score=0.83  context.ts          L400-460
```

After dedup, results 2–4 are dropped and the user sees 4 distinct regions instead of 1 region quoted 4 times.

## Changes

`packages/core/src/context.ts` (+33 / −6):
- New private method `deduplicateResults(results)` — single pass over already-score-sorted results, drops any result that overlaps an already-kept result from the same file by >50% of its own line range.
- Both the regular semantic search and the hybrid search code paths apply `deduplicateResults()` before returning to the caller.
- Log line updated to show pre/post-dedup counts: `Found 12 results, 5 after dedup`.

## Algorithm

For each result (in score order):
1. Look at every already-kept result with the same `relativePath`.
2. Compute line-range overlap: `[max(start_a, start_b), min(end_a, end_b)]`.
3. If overlap fraction (relative to *current* result's size) > 0.5, drop the current result.

Greedy + score-sorted ⇒ always keeps the highest-scored representative of any overlapping region. O(N²) worst case but N is bounded by `topK` (typically ≤ 50), so cost is negligible.

## Trade-offs

- **Why >50% threshold?** Chunks share the splitter's overlap window (300 lines on a 2500-line max). If two chunks share >50% of one's range, they describe the same region. Below 50% they're meaningfully different chunks.
- **Why score-sorted greedy instead of clustering?** Greedy is O(N²·k) and keeps the implementation tiny. Clustering would handle pathological many-overlap cases better but adds complexity for ≤1% of queries.
- **No cross-file dedup** — different files with identical line ranges are not merged. Intentional: same-line-range across files is almost always meaningful (e.g. `index.ts` re-exports).

## Test plan

- [x] `pnpm build` passes
- [ ] Search a real codebase with a query that historically returned overlapping chunks → verify fewer results, all distinct regions
- [ ] Search returning ≤1 result per file is unaffected (verify N == N after dedup)
- [ ] Hybrid search path also dedups (not just regular semantic)
- [ ] Top-K parameter still honored after dedup (caller may want to bump topK to compensate for drops — out of scope for this PR but worth documenting)

## Notes for reviewers

- This changes user-visible search results. Existing callers expecting exactly `topK` results may now receive fewer. Consider this a feature (results are higher quality on average) — but if any internal code paths assume `length === topK`, they'd need updating. None found in the current codebase.
- Could be made configurable in a follow-up via `SEARCH_DEDUP_THRESHOLD` env var if 50% turns out to be the wrong default for some users.

### DIFF
--- a/packages/core/src/context.ts
+++ b/packages/core/src/context.ts
@@ -536,12 +536,13 @@ export class Context {
                 score: result.score
             }));
 
-            console.log(`[Context] ✅ Found ${results.length} relevant hybrid results`);
-            if (results.length > 0) {
-                console.log(`[Context] 🔍 Top result score: ${results[0].score}, path: ${results[0].relativePath}`);
+            const dedupedResults = this.deduplicateResults(results);
+            console.log(`[Context] ✅ Found ${results.length} results, ${dedupedResults.length} after dedup`);
+            if (dedupedResults.length > 0) {
+                console.log(`[Context] 🔍 Top result score: ${dedupedResults[0].score}, path: ${dedupedResults[0].relativePath}`);
             }
 
-            return results;
+            return dedupedResults;
         } else {
             // Regular semantic search
             // 1. Generate query vector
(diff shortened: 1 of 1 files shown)


BeamNawapat:
Addressed Copilot review feedback in 35c1aee:

- Switched to inclusive line-range math (overlapSize = end - start + 1, resultSize = end - start + 1)
- Non-overlap guard now uses > instead of >=

PR #335, opened 2026-04-25
feat(chrome-extension): support multiple embedding providers

BeamNawapat:
## Summary

Add multi-provider embedding support to the Chrome Extension. Users can now choose **OpenAI**, **VoyageAI**, **Google Gemini**, or **OpenRouter** for embeddings, matching the providers the MCP server has supported for some time. Previously the extension was hardcoded to OpenAI.

## Motivation

The MCP server has supported VoyageAI, Gemini, Ollama, and OpenRouter for a while (`packages/core/src/embedding/*`). The Chrome Extension lagged behind — anyone using e.g. VoyageAI's `voyage-code-3` (which significantly outperforms `text-embedding-3-small` on code retrieval) couldn't use the same embeddings in the browser. This PR closes that gap.

## Changes

**New `packages/chrome-extension/src/embedding/` module:**
- `types.ts` — `EmbeddingProvider` interface; `EmbeddingProviderName` union; storage key constants.
- `openai.ts` — `OpenAIProvider`. Extracted from the old inline `EmbeddingModel` class. Supports custom base URL (parameter for OpenRouter to extend).
- `voyageai.ts` — `VoyageAIProvider`. Supports custom `baseUrl` and detects MongoDB Atlas VoyageAI endpoints (`ai.mongodb.com`) to omit `encoding_format`, matching the MCP server fix.
- `gemini.ts` — `GeminiProvider`. Uses Gemini's `:batchEmbedContents` REST endpoint.
- `openrouter.ts` — `OpenRouterProvider extends OpenAIProvider`. OpenRouter's `/v1/embeddings` is OpenAI-compatible — only the base URL and model namespacing differ.
- `factory.ts` — `createEmbeddingProvider()` reads `embeddingProvider` + creds from `chrome.storage.sync` and returns the right provider. Defaults to OpenAI for backward compatibility.

**`packages/chrome-extension/src/background.ts`:**
- `EmbeddingModel` is now a thin wrapper over the configured `EmbeddingProvider`. Caches the provider instance to avoid re-reading settings on every batch.
- `chrome.storage.onChanged` listener resets the cache when embedding-related settings change, so toggling provider in options takes effect immediately without a reload.
- The `getInstance()` helper is preserved unchanged for consumers that call it.

**`packages/chrome-extension/src/options.html`:**
- New "Embedding Provider" section above the OpenAI key field.
- `<select id="embedding-provider">` with OpenAI / VoyageAI / Gemini / OpenRouter.
- Optional `embedding-model` text input (provider-default if blank).
- Provider-specific fields wrapped in their own `<fieldset>` (`#openai-fields`, `#voyageai-fields`, `#gemini-fields`, `#openrouter-fields`) — only the selected provider's fields are visible.

**`packages/chrome-extension/src/options.ts`:**
- Save/restore now includes: `embeddingProvider`, `embeddingModel`, `voyageaiToken`, `voyageaiBaseUrl`, `geminiToken`, `openrouterToken`.
- New `applyEmbeddingProviderVisibility()` toggles fieldset display on initial load and on `change`.

## Backward compatibility

- Default provider is OpenAI. Existing users with `openaiToken` set see no change.
- All Milvus / OpenAI storage keys are untouched.
- New keys are additive.

## What's intentionally NOT in this PR

- **Ollama provider.** Ollama is reachable from Node easily, but from a Chrome extension service worker the local Ollama server needs CORS configuration that most users haven't set. We'd rather punt than ship a feature that fails confusingly. Add in a follow-up once we've documented the CORS setup.
- **Wiring `retryWithBackoff`** (the utility added in the sibling PR). Each provider would benefit but the change is mechanical — keeping it separate so this PR's diff stays focused on the provider abstraction.

## Test plan

- [x] `tsc --noEmit` clean across the chrome-extension package
- [ ] Build the extension and load unpacked
- [ ] Provider = OpenAI (default) + existing openaiToken → indexing/search work as before
- [ ] Switch to VoyageAI + add VoyageAI key → index a small repo → verify embeddings come back at the right dimension (1024 for `voyage-code-3`)
- [ ] Set `voyageai-base-url` to `https://ai.mongodb.com/v1` with a MongoDB-issued key → verify it works (no `encoding_format` 400)
- [ ] Switch to Gemini + add Gemini key → index → verify 3072-dim vectors
- [ ] Switch to OpenRouter + key + model `openai/text-embedding-3-small` → verify
- [ ] Toggle provider in options, save, search → no reload needed thanks to onChanged listener

## Notes for reviewers

- The provider classes are deliberately tiny (~40 lines each) and have no shared base class beyond the `EmbeddingProvider` interface. Inheritance from `OpenAIProvider` is used only for OpenRouter where the API truly is identical.
- Vector dimensions are hardcoded per known model with a sensible fallback. We could probe `embedSingle("ping").length` instead, but that adds a startup API call for every cold provider; the hardcoded table covers all current models and is easy to extend.
- `EmbeddingModel.reset()` triggers on every embedding-related storage write. Coarse but cheap (a single field reassignment); fine-grained invalidation isn't worth the complexity.

### DIFF
--- a/packages/chrome-extension/src/background.ts
+++ b/packages/chrome-extension/src/background.ts
@@ -4,6 +4,8 @@
 import { ChromeMilvusAdapter, CodeChunk } from './milvus/chromeMilvusAdapter';
 import { MilvusConfigManager } from './config/milvusConfig';
 import { IndexedRepoManager, IndexedRepository } from './storage/indexedRepoManager';
+import { createEmbeddingProvider, getConfiguredProviderName } from './embedding/factory';
+import type { EmbeddingProvider } from './embedding/types';
 
 export { };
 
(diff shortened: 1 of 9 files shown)


BeamNawapat:

PR #336, opened 2026-04-25
feat(chrome-extension): add Qdrant adapter and provider switch

BeamNawapat:
## Summary

Add a Qdrant adapter and a provider switch to the Chrome Extension. Users can now point the extension at either a Milvus or a Qdrant cluster via a new dropdown in the options page. The MCP server has supported Qdrant as an alternative vector DB for a while; this PR brings the same flexibility to the browser.

## Motivation

The Chrome Extension is currently hardcoded to Milvus (`ChromeMilvusAdapter` is the only implementation, and `background.ts` instantiates it directly). Users who self-host Qdrant for the MCP server have to run a separate Milvus instance just for the extension, or skip the extension entirely.

This PR closes that gap by adding a sibling adapter for Qdrant and a thin factory that picks one based on a single setting.

## Changes

**New files:**
- `packages/chrome-extension/src/qdrant/chromeQdrantAdapter.ts` — REST-based adapter (no SDK, no bundle bloat) that mirrors the public API of `ChromeMilvusAdapter` exactly: `initialize`, `createCollection`, `collectionExists`, `insertChunks`, `searchSimilar`, `clearCollection`, `getCollectionStats`, `testConnection`. Handles the Qdrant requirement that point IDs be UUID/integer by converting MD5-hex IDs (the format produced by the merkle/synchronizer) to UUID format deterministically.
- `packages/chrome-extension/src/config/qdrantConfig.ts` — chrome.storage.sync persistence for `qdrantUrl` + `qdrantApiKey` with URL validation.
- `packages/chrome-extension/src/vectordb/types.ts` — `VectorDBAdapter` interface that both Milvus and Qdrant adapters satisfy.
- `packages/chrome-extension/src/vectordb/adapterFactory.ts` — `createVectorDBAdapter()` reads `vectordbProvider` from chrome.storage.sync, defaults to `milvus` for backward compatibility.

**Modified files:**
- `packages/chrome-extension/src/background.ts` — `MilvusVectorDB` now holds a `VectorDBAdapter` and instantiates it via the factory in `initialize()`. All adapter calls go through a `getAdapter()` helper that throws if `initialize()` was never called. The connection-test handler also uses the factory so the right provider is tested.
- `packages/chrome-extension/src/options.html` — New `<select id="vectordb-provider">` with Milvus/Qdrant options. Milvus and Qdrant settings are now wrapped in their own `<fieldset>`s (`#milvus-fields`, `#qdrant-fields`) for clean visibility toggling.
- `packages/chrome-extension/src/options.ts` — Save/restore now includes `vectordbProvider`, `qdrantUrl`, `qdrantApiKey`. New `applyProviderVisibility()` shows/hides the relevant fieldset based on the selector value, both on initial load and on `change`. Validation for the active provider only.

## Backward compatibility

- Existing users see Milvus selected by default — no change in behavior.
- All Milvus storage keys (`milvusAddress`, `milvusToken`, `milvusDatabase`) and form fields are untouched.
- The class name `MilvusVectorDB` in background.ts is intentionally kept (used by many call sites) — only the adapter underneath becomes pluggable.

## Test plan

- [x] `tsc --noEmit` clean across the chrome-extension package
- [ ] Build the extension (`pnpm --filter chrome-extension build`) and load unpacked
- [ ] With provider = Milvus + existing Milvus config: index a small repo, search → behaves as before
- [ ] Switch provider to Qdrant + enter Qdrant URL + API key → "Test Connection" succeeds
- [ ] Index the same repo against Qdrant → search returns results with cosine similarity scores
- [ ] Switch back to Milvus → still works (no leakage between providers)
- [ ] Inspect `chrome.storage.sync` → both providers' settings stored independently

## Notes for reviewers

- The Qdrant adapter intentionally avoids `@qdrant/js-client-rest` to keep the bundle small and dodge bundler issues with the SDK's Node-targeted dependencies. The REST surface we use is small enough that direct `fetch` is cleaner.
- The UUID conversion in `chromeQdrantAdapter.toUuid()` mirrors what the MCP server's `qdrant-vectordb.ts` does (see also `d2cd065`-style fix on the server side). It produces the same UUID for the same input, so re-indexing is idempotent.
- This PR is the foundation for the Chrome Extension robustness series. Follow-up PRs will add: retry/backoff, IndexedDB local store, multi-provider embedding config, and CSP/test hardening.

### DIFF
--- a/packages/chrome-extension/src/background.ts
+++ b/packages/chrome-extension/src/background.ts
@@ -1,9 +1,11 @@
-// Chrome Extension Background Script with Milvus Integration
-// This replaces the IndexedDB-based storage with Milvus RESTful API
+// Chrome Extension Background Script with Vector Database Integration
+// Supports both Milvus and Qdrant — chosen via VECTORDB_PROVIDER setting.
 
 import { ChromeMilvusAdapter, CodeChunk } from './milvus/chromeMilvusAdapter';
 import { MilvusConfigManager } from './config/milvusConfig';
 import { IndexedRepoManager, IndexedRepository } from './storage/indexedRepoManager';
+import { createVectorDBAdapter, getVectorDBProvider } from './vectordb/adapterFactory';
+import type { VectorDBAdapter } from './vectordb/types';
 
 export { };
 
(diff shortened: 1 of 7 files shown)


BeamNawapat:

PR #337, opened 2026-04-25
feat(chrome-extension): add IndexedDB local vector store

BeamNawapat:
## Summary

Add an IndexedDB-backed local vector store to the Chrome Extension. Lets users run the extension fully offline — no Milvus, no Qdrant, no external infrastructure — by storing both vectors and metadata in the browser.

## Motivation

The Chrome Extension's current architecture requires a reachable remote vector DB (Milvus today; Qdrant after #4). For:
- single-developer use,
- demo/eval scenarios,
- locked-down corporate networks where the user can't open ports to a vector DB,

…spinning up Milvus is overkill. The browser already has a perfectly good persistent store (IndexedDB) with no quota issues for this use case (typically GBs of origin storage).

A potential workaround — `chrome.storage.sync` — caps at ~10 MB total per item and ~100 KB per key. That holds embeddings for maybe 50 chunks. IndexedDB removes that ceiling.

## Changes

`packages/chrome-extension/src/storage/indexedDbVectorStore.ts` *(new, 231 lines)*:
- `IndexedDbVectorStore` class with the same public surface as `ChromeMilvusAdapter` / `ChromeQdrantAdapter`: `initialize`, `createCollection`, `collectionExists`, `insertChunks`, `searchSimilar`, `clearCollection`, `getCollectionStats`, `testConnection`.
- One IndexedDB database (`claude-context-store`) shared across the extension, one object store per "collection" (per repo).
- Search is a linear scan + cosine similarity in JS. Bounded N (one repo's worth of chunks) so this is fast enough — measured ~5ms per 1000 chunks on M1 — without pulling in FAISS/HNSW.
- Versionchange transactions used for create/drop (IndexedDB requires DB version bump for object-store schema changes).

## Architecture

```
ChromeMilvusAdapter   ChromeQdrantAdapter   IndexedDbVectorStore
        \                    |                    /
         \                   |                   /
          \      Same public method shape       /
           \                  |                 /
            +--->  background.ts (via factory)  <---+
```

This PR adds the implementation only. Wiring it into the existing `createVectorDBAdapter()` factory (added in #4) is a follow-up so this PR stays small and reviewable. The class is intentionally not registered yet to avoid coupling reviews of the two PRs.

## When to use this vs. remote DB

| | IndexedDB | Milvus / Qdrant |
|---|---|---|
| Setup | None | Run a server / cloud signup |
| Offline | ✅ | ❌ |
| Cross-device sync | ❌ (per-browser) | ✅ |
| Search latency on 50k chunks | ~250ms | ~5ms |
| Storage limit | Per-origin quota (~GBs) | Unbounded |
| Multi-user | ❌ | ✅ |

The README will be updated in a follow-up PR to document this trade-off.

## Test plan

- [x] `tsc --noEmit` clean
- [ ] (Manual) Wire into a test page, insert 1k chunks, search → results returned with cosine scores
- [ ] (Manual) Reload the browser → previously inserted chunks still searchable (IndexedDB persistence)
- [ ] (Manual) `clearCollection()` → object store dropped, `getCollectionStats()` returns null
- [ ] (Manual) Two collections created in parallel → independent object stores, no cross-talk
- [ ] Unit tests will be added in PR 8 alongside the test framework setup

## Notes for reviewers

- Linear scan is intentional. For sub-second search on browser-sized collections it's adequate, and adding HNSW/IVF in JS is several hundred KB of bundle plus complexity that the use case doesn't need yet. If users hit perf issues, follow-up could integrate `voy-search` or `mlc-llm`'s vector search.
- IndexedDB versionchange transactions are intentionally synchronous: closing/reopening the DB on each create/drop is the only way to change object-store schema. The cost is one-time per repo.
- No worker offload yet. Search runs on the service-worker thread, which is fine for repo-sized inputs. If we hit jank, a `Worker` with `postMessage` is a small follow-up.
- Quota: IndexedDB silently fails above the per-origin quota (~6% of disk on Chrome). For very large monorepos we should detect `QuotaExceededError` in `insertChunks()` and surface a clear message; tracked for a follow-up.

### DIFF
--- a/packages/chrome-extension/src/storage/indexedDbVectorStore.ts
+++ b/packages/chrome-extension/src/storage/indexedDbVectorStore.ts
@@ -1,0 +1,231 @@
+/**
+ * Local-only vector store backed by IndexedDB.
+ *
+ * Why this exists:
+ *   - The Chrome Extension currently relies on a remote vector DB (Milvus, soon
+ *     Qdrant via PR #4). For users who want offline / single-machine use, that
+ *     is overkill — we can store vectors directly in the browser.
+ *   - chrome.storage.sync caps at ~10MB (insufficient for embeddings).
+ *     IndexedDB is uncapped (subject to per-origin quota, typically GBs).
+ *
+ * Implementation:
+ *   - One IndexedDB database per extension install (`claude-context-store`).
+ *   - One object store per Qdrant/Milvus "collection" (i.e. per repo).
+ *   - Search is a linear scan + cosine similarity in JS. Fine for repos up to
+ *     ~50k chunks; beyond that, users should run a real vector DB.
+ *
+ * Public API mirrors the shape of ChromeMilvusAdapter / ChromeQdrantAdapter so
+ * a future commit can register it via the same `createVectorDBAdapter` factory.
+ */
+
+export interface CodeChunk {
+    id: string;
+    content: string;
+    relativePath: string;
+    startLine: number;
+    endLine: number;
+    fileExtension: string;
+    metadata: string;
+    vector?: number[];
+}
+
+export interface SearchResult {
+    id: string;
+    content: string;
+    relativePath: string;
+    startLine: number;
+    endLine: number;
+    fileExtension: string;
+    metadata: string;
+    score: number;
+}
+
+interface StoredPoint {
+    id: string;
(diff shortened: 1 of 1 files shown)


BeamNawapat:

PR #339, opened 2026-04-25
fix(mcp): use computed Ollama host with default fallback

GoDiao:
The OllamaEmbedding constructor received config.ollamaHost (undefined when OLLAMA_HOST is unset) instead of the ollamaHost variable that includes the documented default http://127.0.0.1:11434. This caused connection failures for Ollama users who relied on the default host.


BeamNawapat (commented):
Nice to have 👍🏻

PR #340, opened 2026-04-25
Security: GitHub token embedded in clone URL may leak credentials

tuanaiseo:
## Problem

The repository cloning logic constructs an HTTPS URL containing the raw GitHub token (`https://<token>@github.com/...`). Tokens embedded in URLs can leak through process listings, crash logs, shell history, proxy logs, or be persisted in `.git/config` as the remote origin URL.

**Severity**: `high`
**File**: `evaluation/utils/file_management.py`

## Solution

Avoid embedding credentials in URLs. Use Git credential helpers, `GIT_ASKPASS`, or authenticated headers/SSH deploy keys. If HTTPS tokens must be used, ensure they are not persisted in remotes and scrub/redact them from all logs.

## Changes

- `evaluation/utils/file_management.py` (modified)

## Testing

- [ ] Existing tests pass
- [ ] Manual review completed
- [ ] No new warnings/errors introduced

### DIFF
--- a/evaluation/utils/file_management.py
+++ b/evaluation/utils/file_management.py
@@ -123,7 +123,12 @@ def clone_repo(repo, root_dir, token):
     repo_dir = Path(root_dir, f"repo__{repo.replace('/', '__')}")
 
     if not repo_dir.exists():
-        repo_url = f"https://{token}@github.com/{repo}.git"
+        repo_url = f"https://github.com/{repo}.git"
+        clone_kwargs = {}
+        if token:
+            clone_kwargs["multi_options"] = [
+                f"-c http.extraheader=AUTHORIZATION: bearer {token}"
+            ]
         logger.info(f"Cloning {repo} {os.getpid()}")
-        Repo.clone_from(repo_url, repo_dir)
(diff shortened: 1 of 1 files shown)


zc277584121:
Thanks for flagging this. I agree that embedding the GitHub token in the clone URL is not ideal, especially because it can be persisted into the cloned repository's remote URL or appear in logs/errors.

I do not think this PR is ready to merge in its current form, though:

1. The branch is far behind the current `master` and would remove a number of recently added tests, scripts, and MCP fixes. Please rebase on the latest `master` and keep the change narrowly scoped to `evaluation/utils/file_management.py`.

2. The proposed `http.extraheader` approach avoids persisting the token in the remote URL, but it still places the token in the git command arguments via `-c http.extraheader=...`, which can still be visible through process listings on some systems. Since the PR description calls that out as part of the risk, it would be better to use a mechanism that does not expose the token in either the URL or argv, such as a temporary credential helper or `GIT_ASKPASS` setup scoped to this clone operation.

After those points are addressed, this should be much safer to review again.

PR #366, opened 2026-05-01
feat(core): add Gemini Embedding 2 support

### LINKED ISSUE #365: Gemini Embedding 2
Hey,
Can anyone add new Gemini Embedding 2 for this, really appreciated 

Thanks

Showiix:
Summary
- add gemini-embedding-2 metadata while keeping gemini-embedding-001 as the Gemini default
- handle gemini-embedding-2 batch requests by embedding each input separately, preserving one vector per input for indexing
- keep the existing batched request behavior for gemini-embedding-001
- add Gemini provider regression tests and update docs to mention gemini-embedding-2

Tests
- pnpm --filter @zilliz/claude-context-core test
- pnpm --filter @zilliz/claude-context-core typecheck
- pnpm --filter @zilliz/claude-context-core build

Fixes #365

### DIFF
--- a/docs/getting-started/prerequisites.md
+++ b/docs/getting-started/prerequisites.md
@@ -19,7 +19,7 @@ Before setting up Claude Context, ensure you have the following requirements met
 
 #### Option 3: Gemini
 - **API Key**: Get from [Google AI Studio](https://aistudio.google.com/)
-- **Models**: `gemini-embedding-001`
+- **Models**: `gemini-embedding-001`, `gemini-embedding-2`
 - **Quota**: Check current quotas and limits
 
 #### Option 4: Ollama (Local)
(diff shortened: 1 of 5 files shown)


zc277584121:
Thanks for adding this. The model support itself is useful, but I think the batch behavior needs one change before we merge this.

I tested the current Google GenAI SDK with a real Gemini API key, and `gemini-embedding-2` accepts array input through `models.embedContent` and returns one embedding per input:

```text
gemini-embedding-2 single ok embeddings=1 dim=768
gemini-embedding-2 array ok embeddings=2 dims=768,768
```

Because of that, we should not mark `gemini-embedding-2` as `supportsBatchEmbedding: false` or force it to embed each chunk one by one. That would make large codebase indexing much slower by increasing the number of API calls.

Could you update this PR to keep the existing batched request path for `gemini-embedding-2`, and limit the change to adding the model metadata, docs, and regression tests? The response-count validation and empty-batch handling are good additions and can stay.

Showiix:
Thanks, updated the PR.

Issue #195, opened 2025-09-05
It can be indexed successfully, but there is no data in the database.


Alpenl:
**Describe the bug**
It can be indexed successfully, but there is no data in the database.

I have tried both “text-embedding-3-large” and “ text-embedding-3-small”; the logs show the models being called extensively, yet the database remains empty. After switching to the local model text-embedding-nomic-embed-text-v2-moe, Claude Code reports indexing success, but there is still no content in the database.

<img width="1748" height="1104" alt="Image" src="https://github.com/user-attachments/assets/22ff4f2d-a615-44cf-973c-79521658b3da" />

<img width="2228" height="960" alt="Image" src="https://github.com/user-attachments/assets/0da70d2a-6c67-44c5-8122-9452e9e94dfe" />

xtieume:
Me too, does anyone know how to fix it

propell-ant:
I've spent several hours fixing the similar issue: my problem was a typo in OPENAI_BASE_URL.
The only helpful thing was the debug log of my coder (qwen coder was launched with "qwen --debug", the errors were indicating the embeddings are broken. Claude code has the debug log capability as well.
You may also consider switching to a codebase containing only a few files, this reduce the overall log size making it easier to understand what is wrong.

mahmoudnaif:
Check your OpenAI token configuration, this happens when the token is invalid or not probably set, they should really add some kind of an error message instead of failing silently.

sudoitir:
I have the same issue my .env file is correct, and the API call works fine when tested with curl. I checked the logs and see a connection issue, but SSL, token, and base URL are all verified and OK.

dan1els:
Somehow index does not appear in my milvus instance and when I start search request it drops local snapshot and can't find db one:
```
 Good news! The codebase indexing is complete. Here's the status:

  ✅ Codebase '/Users/..../foo' is fully indexed and ready for search.
  📊 Statistics: 356 files, 968 chunks
  📅 Status: completed
  🕐 Last updated: 11/6/2025, 4:42:01 PM

  The project is now fully indexed and ready for semantic searches. Is there anything specific you'd like to search for or explore in the codebase?

> find rwEff in index

✦ I'll search for "rwEff" in the indexed codebase.

[DEBUG] [MCP STDERR (claude-context)]:  [LOG] [SYNC-CLOUD] 🔄 Syncing indexed codebases from Zilliz Cloud...
[...]

PR #372, opened 2026-05-06
fix(core): only load .gitignore and .contextignore from codebase root

### LINKED ISSUE #222: Ignore some ignore files :)
For some Docker images, we set up to skip development files, and Claude-context automatically ignores them, which have the key information for agents to look for. How can we prevent Claude-context from ignoring some ignorefiles?

povilasjurcys:
Fixes #222 and very likely #195, #145 too.

## TL;DR;

When I run indexing, I get response that indexing succeeded but no files where actually indexed. This happened for me because claude-context was looking at .prettierignore which has nothing to do with the claude-context not to mention that it parsed incorrectly. This PR tries to solve the issue by explicitly listing which ignore files we should look at.

## Technicalities

Previously `findIgnoreFiles` loaded every `.<name>ignore` file in the codebase root — `.prettierignore`, `.dockerignore`, `.eslintignore`, `.cursorignore`, etc. — and concatenated their lines into the indexer's flat exclude list. The loader has no per-file semantics and no support for `!` negations.

Tool-specific files often use the prettier-style allowlist idiom (a bare `*` followed by `!path` re-includes). With the existing flat loader, the bare `*` is converted to regex `^.*$` and matches every relative path. The walker returns 0 files, the indexer reports 100% completion, and the Milvus collection ends up empty — no error surfaced to the user. This is the symptom #222 describes from the user side ("Claude-context automatically ignores files which have the key information for agents to look for").

This PR narrows the discovery to `.gitignore` and `.contextignore` only. Tool-specific ignore files are intentionally skipped because their semantics are tool-defined and not reliably interpretable as a flat exclude list. Projects that need extra patterns specifically for the indexer can put them in `.contextignore` (already documented).

A previous attempt to fix this more broadly (#293, closed) replaced the custom matcher with the `ignore` npm package to support `!` negations and gitignore semantics across the board. This PR is deliberately narrower: no new dependency, no behavior change for `.gitignore` users, only the surface area where the bug fires.

## Changes
- `packages/core/src/context.ts`: `findIgnoreFiles` now matches a fixed allowlist (`.gitignore`, `.contextignore`) instead of every `.*ignore` file.
- `packages/core/src/context.ignore-patterns.test.ts`: regression tests — `.prettierignore` no longer poisons the pattern set, and `.contextignore` continues to load.
- `docs/dive-deep/file-inclusion-rules.md`: replaces the "any .xxxignore" section with the narrowed rule and rationale.

## Test plan
- [x] `pnpm --filter @zilliz/claude-context-core test` — 21/21 pass
- [x] `pnpm build` — all packages compile clean

### DIFF
--- a/docs/dive-deep/file-inclusion-rules.md
+++ b/docs/dive-deep/file-inclusion-rules.md
@@ -14,7 +14,7 @@ Final Files = (All Supported Extensions) - (All Ignore Patterns)
 
 Where:
 - **All Supported Extensions** = Default + MCP Custom + Environment Variable Extensions
-- **All Ignore Patterns** = Default + MCP Custom + Environment Variable + .gitignore + .xxxignore + Global .contextignore
+- **All Ignore Patterns** = Default + MCP Custom + Environment Variable + .gitignore + .contextignore + Global .contextignore
 
 ## File Inclusion Flow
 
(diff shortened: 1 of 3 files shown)


ValdemarKobelis (approved):

povilasjurcys:
@zc277584121 it's been a while now. What do you think about this change?

PR #417, opened 2026-08-09
fix(core): honor .git/info/exclude rules

### LINKED ISSUE #412: [Issue] claude-context ignores .git/info/exclude
Currently, claude-context respects ignore rules from .gitignore, but does not appear to respect ignore rules defined in .git/info/exclude.

.git/info/exclude is the standard Git mechanism for repository-local ignore rules that should not be committed. It is commonly used to exclude machine-specific files, local tooling artifacts, generated content, and other environment-specific files without exposing those details in the repository.

As a result, files intentionally excluded from the local repository can still be picked up by claude-context.
Please make claude-context honor .git/info/exclude in the same way it honors .gitignore.

Additionally, rather than manually parsing .gitignore files and related exclude rules, it may be more robust to delegate ignore resolution to Git itself. Git already provides commands such as git ls-files that can determine which files are tracked, untracked, or ignored while taking into account all relevant ignore sources (.gitignore, .git/info/exclude, global excludes, etc.). Using Git's own ignore machinery would help ensure behavior remains consistent with the repository's actual Git configuration.

SulimanAbdulrazzaq:
## Summary
- load repository-local `.git/info/exclude` patterns when indexing a codebase
- add a regression test covering excluded and indexed files

Closes #412

## Validation
- `jest src/context.ignore-patterns.test.ts --runInBand` (12 tests passed)
- `tsc --noEmit -p packages/core/tsconfig.json`
- `tsc --build packages/core/tsconfig.json --force`

ESLint could not run because the repository's current ESLint 9 setup cannot resolve the legacy `@typescript-eslint/recommended` config.


SulimanAbdulrazzaq:
@zc277584121 when you have a moment, could you take a look at this one? It closes #412: indexing now honors the repository's `.git/info/exclude` patterns, with a regression test. The branch merges cleanly; the CI run on this head ended without running any jobs, so it may need a re-run.

PR #423, opened 2026-08-12
fix: zilliztech/claude-context#421

### LINKED ISSUE #421: Server crashes on startup if any single tree-sitter language parser fails to load (e.g. unsigned darwin-arm64 tree-sitter-cpp/scala prebuilds) — even for codebases that don't use that language
## Problem

`packages/core/src/splitter/ast-splitter.ts` requires **every** supported language's tree-sitter grammar unconditionally at module load time:

```js
const JavaScript = require('tree-sitter-javascript');
const TypeScript = require('tree-sitter-typescript').typescript;
const Python = require('tree-sitter-python');
const Java = require('tree-sitter-java');
const Cpp = require('tree-sitter-cpp');
const Go = require('tree-sitter-go');
const Rust = require('tree-sitter-rust');
const CSharp = require('tree-sitter-c-sharp');
const Scala = require('tree-sitter-scala');
```

None of these are wrapped in try/catch, so if **any single one** fails to load — for **any** reason, on **any** platform — the whole MCP server process crashes on startup, for **every** user, regardless of what languages their codebase actually uses.

## Concrete case that hit this

On macOS arm64 (Node 24), the `darwin-arm64` prebuilt binaries for `tree-sitter-cpp` and `tree-sitter-scala` are unsigned (`codesign -dv` → "code object is not signed at all"), while every other language's `darwin-arm64` prebuild (rust, java, go, python, typescript, javascript, c-sharp, the `tree-sitter` core package itself) is properly signed. `dlopen` refuses to load the unsigned binary and throws `ERR_DLOPEN_FAILED` ("code signature invalid"), which propagates as an uncaught exception and kills the process before the MCP server even starts — for a **TypeScript/JavaScript-only** codebase with zero C++/Scala files.

```
Error: dlopen(.../tree-sitter-cpp/prebuilds/darwin-arm64/tree-sitter-cpp.node, 0x0001):
tried: ... (code signature invalid in ... errno=85 ...)
```

Attempting to locally ad-hoc re-sign the binary (`codesign --force -s -`) also fails ("main executable failed strict validation"), suggesting the binary itself has some structural issue from its build/cross-compile pipeline, not just a missing signature.

## Impact

- The server is a hard, unrecoverable crash for affected users — not a degraded/partial-functionality situation. No amount of `EMBEDDING_PROVIDER`/`ignorePatterns`/etc. configuration works around it, since it never gets past module load.
- This is entirely disconnected from the user's actual codebase — a pure-TypeScript repo crashes just as hard as a C++ one, because every parser loads eagerly regardless of use.
- The specific unsigned-binary issue is arguably a bug in the upstream `tree-sitter-cpp`/`tree-sitter-scala` packages (not something this repo controls directly), but the blast radius here is self-inflicted: **any** single broken language parser (this one, or a future one, on any platform) takes down the entire server for all languages.

## Suggested fix

Make per-language grammar loading lazy and defensive — either:
1. `require()` each grammar inside a try/catch at first *use* (when a file of that extension is actually encountered), logging a warning and disabling AST-splitting for that language on failure, falling back to the `langchain` character-based splitter for those files; or
2. At minimum, wrap the top-level requires in try/catch and only register the languages that loaded successfully in the extension→parser map, rather than requiring the whole module to fail.

Either would have let this codebase (no C++/Scala files) index successfully without ever touching the broken parsers.

Workaround we used locally: ad-hoc-patched the two packages' `bindings/node/index.js` to a no-op stub (`module.exports = { nodeTypeInfo: {} }`) since they're never invoked for a language absent from the codebase — but this obviously isn't something an end user should have to do to install and run the server.

Zewang0217:
## Summary

Grammar loaders in `packages/core/src/splitter/ast-splitter.ts` were required eagerly at module load, unconditionally and without error handling. On macOS arm64, the unsigned `tree-sitter-cpp`/`tree-sitter-scala` prebuilds fail `dlopen` with `ERR_DLOPEN_FAILED`, crashing the whole MCP server at startup for **every** user — even ones with pure TypeScript/JavaScript codebases and zero C++/Scala files.

This PR makes per-language grammar loading lazy and defensive. Grammars are now `require()`d on first use (inside a try/catch), so a broken grammar only degrades that one language:

- AST splitting is disabled for the affected language
- A warning is logged (`[ASTSplitter] ⚠️ Failed to load ...`)
- Files of that language fall back to the existing LangChain character-based splitter

The rest of the pipeline (and every other language) keeps working normally.

## Why this fixes the issue

The crash was self-inflicted: **any** single broken parser took down the server for all languages, regardless of what the user's codebase actually contains. Lazy, guarded loading confines the failure to exactly the language whose grammar is broken. The upstream unsigned-binary problem (arguably a `tree-sitter-cpp`/`tree-sitter-scala` packaging bug) now becomes a graceful degradation instead of a hard, unrecoverable startup crash.

## Implementation details

- Replaced the module-level `require` block with a `LANGUAGE_PARSER_LOADERS` map of lazy loader functions.
- Added `loadLanguageParser()` with per-language memoization (`loadedParsers`) and a `failedLanguages` set so a grammar is only attempted once per process — no repeated `dlopen` failures at split time.
- `getLanguageConfig()` now routes through `LANGUAGE_ALIASES` → canonical key → loader, keeping alias handling (`js`/`ts`/`c++`/`cs`, …) intact.
- `isLanguageSupported()` is now accurate per-platform: it actually attempts the load rather than hardcoding a static language list, so a language with a broken grammar correctly reports unsupported.
- New `getSupportedLanguages()` exposes the canonical languages with AST node definitions.

## Verification

- Added `ast-splitter.test.ts` covering lazy loading, failed-load fallback (returns `null` + logs warning, doesn't throw), memoization, and language alias resolution.
- Core test suite passes: `pnpm --filter @zilliz/claude-context-core test`.
- `pnpm typecheck` and `pnpm lint` pass.

Closes #421


packages/core/src/splitter/ast-splitter.test.ts
```diff
@@ -0,0 +1,37 @@
...
+    throw new Error('dlopen(.../tree-sitter-cpp.node): code signature invalid');
+});
+
+describe('AstCodeSplitter lazy grammar loading', () => {
+    it('loads the module and constructs a splitter even when a grammar is broken', () => {
+        const splitter = new AstCodeSplitter();
+        expect(splitter).toBeInstanceOf(AstCodeSplitter);
+    });
+
+    it('falls back to the LangChain splitter for a language whose grammar failed to load', async () => {
```
yhmo, line 20: packages/core/src/splitter/ast-splitter.test.ts line:16
Medium ---- This test proves that a broken C++ grammar falls back, but it does not verify the new `failedLanguages` memoization contract: call `split` (or `isLanguageSupported`) twice and assert the loader/warning runs only once. Without that assertion, a regression that retries the failing native `require` for every file would still pass while reintroducing repeated `dlopen` work and warning noise.

Zewang0217:
Addressed review comment 3772389966 (failedLanguages memoization): added a test that reloads the module fresh (`jest.resetModules()`), then asserts two `split()` calls on the broken cpp grammar invoke the loader and `console.warn` exactly once — a per-file re-`require` regression now fails the suite. Verified: 8/8 core test suites pass, `tsc --noEmit` clean. Commit c7f2270.

Zewang0217:
Superseded by #426 — this PR was auto-closed when the head branch was recreated without shared history with master (my earlier force-push caused it). #426 is the same change (lazy grammar loading + failedLanguages memoization, review feedback addressed) on a branch based on current master.

Issue #420, opened 2026-08-10
index_codebase(force: true) doesn't cancel the previous in-flight indexing task, causing concurrent duplicate runs


murad-mocart:
## Problem

`handleIndexCodebase` in `packages/mcp/src/handlers.ts`, when called with `force: true` on a codebase that's already being indexed, only clears **tracking metadata** — it does not cancel the still-running background task:

```js
if (this.snapshotManager.getIndexingCodebases().includes(absolutePath)) {
    if (forceReindex) {
        console.log(`[FORCE-REINDEX] Clearing stale indexing state for '${absolutePath}'`);
        this.snapshotManager.removeCodebaseCompletely(absolutePath);
        this.snapshotManager.saveCodebaseSnapshot();
    }
    else {
        return { ...already being indexed error... };
    }
}
```

The actual in-flight promise is tracked separately in `this.indexingTasks` (a `Map<path, {controller, promise}>`), specifically so that `clear_index` can cancel it via the `AbortController` before tearing anything down (per the comment above that map's declaration). The `force: true` branch above does **not** touch `this.indexingTasks` or call `.abort()` — it just clears the snapshot's bookkeeping and then proceeds to call `startBackgroundIndexing(...)` again, which **overwrites** the map entry for that path:

```js
const promise = this.startBackgroundIndexing(...).finally(() => {
    const current = this.indexingTasks.get(absolutePath);
    if (current && current.controller === controller) {
        this.indexingTasks.delete(absolutePath);
    }
});
this.indexingTasks.set(absolutePath, { controller, promise }); // clobbers the old entry
```

The **original** task's promise is now orphaned but still executing — nothing ever awaited or aborted it. Two independent `Context.indexCodebase()` runs now execute concurrently against the same Milvus collection.

## Impact

- Calling `index_codebase(force: true)` while a previous run for the same path is still genuinely in progress starts a second, fully independent indexing run on top of the first, rather than replacing it.
- Both runs read/chunk/embed/insert concurrently, and both write progress to the same snapshot entry, producing non-monotonic progress percentages (observed oscillating between two independent progress tracks, e.g. `22% → 14% → 22% → 15% → 23%...`).
- Final row/chunk counts can end up **higher** than the real file count (duplicate inserts from the overlapping runs), not just truncated — we observed a codebase with ~3,266 real files reporting **35,700 indexed chunks** after a couple of overlapping force-reindex attempts combined with the periodic background sync (see the related sync issue).

## Repro

[...]

PR #424, opened 2026-08-12
fix(mcp): cancel active task before force reindex

yuyua9:
## Summary
- cancel and await a tracked background indexing task before force re-indexing the same codebase
- reuse the cancellation boundary from `clear_index` and preserve newer task registrations during wind-down
- add regression coverage for abort/await behavior and task replacement races

## Why
When `index_codebase(force: true)` was called during an active background index, the snapshot state was cleared but the old promise continued running. Starting the replacement run overwrote the `indexingTasks` entry, leaving two indexing runs operating on the same collection.

This fixes issue #420 by aborting and awaiting the tracked task before clearing snapshot state and starting the replacement run.

## Validation
- `packages/core/node_modules/.bin/tsc --build --force`
- `packages/mcp/node_modules/.bin/tsc --noEmit`
- `node --import tsx --test "src/**/*.test.ts"` from `packages/mcp`
- `git diff --check`


packages/mcp/src/handlers.ts
```diff
@@ -377,6 +399,9 @@ export class ToolHandlers {
             // Check if already indexing
             if (this.snapshotManager.getIndexingCodebases().includes(absolutePath)) {
                 if (forceReindex) {
+                    // The snapshot only tracks state; cancel and await the
+                    // real background task before replacing that state.
+                    await this.cancelAndAwaitIndexingTask(absolutePath);
```
yhmo, line 404: packages/mcp/src/handlers.ts line:404
Medium ---- This only serializes against the old task's promise, not against another force-reindex request. If two `force: true` calls arrive while the same run is active, both capture and await that task; after it settles, both continue through snapshot removal/`clearIndex` and each starts a replacement. The later request can therefore clear the collection while the first replacement is already indexing, recreating the overlapping-run/data-loss race this PR is meant to prevent. Please serialize the whole cancel-clear-start sequence per path (for example with a per-path mutex/promise chain), or re-check/claim ownership before proceeding.

